PR #5, opened 2020-02-05
Skip Calico-kube-controllers deployment for clusters without a backend

zanetworker:
This PR removes `calico-kube-controllers` components for clusters running without a backend (e.g., Azure). This will further improve our availability monitoring as we won't get alerts related to unused components.

**Release note**:

```improvement operator
Clusters without a backedend (i.e., backend is set to `None`) will run without the unneeded `calico-kube-controllers` deployment
```

### DIFF
--- a/charts/internal/calico/templates/kube-controller/clusterrole-calico-kube-controllers.yaml
+++ b/charts/internal/calico/templates/kube-controller/clusterrole-calico-kube-controllers.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.config.kubeControllers.enabled }}
 ---
 kind: ClusterRole
 apiVersion: {{ include "rbacversion" . }}
@@ -43,3 +44,4 @@ rules:
       - get
       - create
       - update
+{{- end }}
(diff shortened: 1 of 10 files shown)


timuthy:
What do you think about adding all calico-kube-controller related manifests to a sub-chart and control the deployment via a `requirements.yaml`?
Example: https://github.com/gardener/gardener/blob/master/charts/shoot-addons/requirements.yaml

zanetworker:
@timuthy thanks for the feedback :) 

While I think it makes sense for add-ons as an example, I am not sure it similarly fits here, `calico-kube-controllers` or `typha` are not whole components on their own compared to Add-ons (e.g., Dashboard), so I can not completely apply the concept of subchart for the entire Calico deployment. If we do that the `requirements.yaml` would simply consist of two components and the rest will stay as folders. The division of folder structure here is for convenience in my opinion rather than hierarchy. In Gardener main this is the same for our main charts (https://github.com/gardener/gardener/tree/master/charts).

timuthy:
@zanetworker valid points. I just wanted to avoid a lot of `if enabled` cases.

PR #12, opened 2020-03-11
backport calico charts, most important adopt calico-node liveness pro…

majst01:
…be execution, nodeselector now out of beta

We found some differences between the charts shipped here and upstream calico. I changed:
- calico-node liveness probe
- calico-node got one more rbac rule
- enlarge calico-node cpu requests

```

### DIFF
--- a/charts/internal/calico/templates/kube-controller/deployment-calico-kube-controllers.yaml
+++ b/charts/internal/calico/templates/kube-controller/deployment-calico-kube-controllers.yaml
@@ -32,7 +32,7 @@ spec:
         gardener.cloud/role: system-component
     spec:
       nodeSelector:
-        kubernetes.io/os: linux
+        beta.kubernetes.io/os: linux
       tolerations:
         # Mark the pod as a critical add-on for rescheduling.
         - key: CriticalAddonsOnly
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: charts/internal/calico/templates/kube-controller/deployment-calico-kube-controllers.yaml [hunk 1]
```diff
@@ -32,7 +32,7 @@ spec:
         gardener.cloud/role: system-component
     spec:
       nodeSelector:
-        beta.kubernetes.io/os: linux
+        kubernetes.io/os: linux
```
rfranzke: Will this work on 1.10 clusters?

majst01: 1.10 ? still exists, unsure, will revert this if 1.10 is required to work as well

majst01: reverted

zanetworker:
Hi @majst01, 

I have a few questions: 

- why do we need to support kubeadm use-cases? 
- by upstream you mean the master branch or 3.11.2?

majst01:
No need to suport kubeadm, but i feel it difficult to cherry pick parts of the charts, instead of using complete upstream. 
I compared against 3.13 most recent stable.

zanetworker:
@majst01 currently we support 3.11.2, thats why the config is different.

Does it make sense to file a different PR that upgrades calico to 3.13, which also includes updating the images to comply with the YAMLs?

majst01:
Yes sure, i was a bit afraid of doing so in one go, but i can do this.

majst01:
Hi @zanetworker new PR is https://github.com/gardener/gardener-extension-networking-calico/pull/20

PR #14, opened 2020-03-13
Increase calico-node limits

zanetworker:
**What this PR does / why we need it**:
Increase the limits for calico-node daemonset. 

cc @ialidzhikov 

**Which issue(s) this PR fixes**:
Fixes #13 

**Special notes for your reviewer**:

**Release note**:

```improvement operator
Calico requests and limits are now increased to avoid resource starvation. 
```

PR #21, opened 2020-03-20
raise calico-node cpu limits to survive setups without bird and ~150+…

majst01:
… pods per node

Running calico-node in a setup without bird, 800m cpu limit is hit quite often and keeps lots of conntrack commands stay stuck for ~20sec. This brings the calico-node pod in unhealthy state as healtcheck and liveliness check will not succeed.

Can we use this PR just to create the image please ?

### DIFF
--- a/charts/internal/calico/templates/node/daemonset-calico-node.yaml
+++ b/charts/internal/calico/templates/node/daemonset-calico-node.yaml
@@ -178,7 +178,7 @@ spec:
               cpu: 250m
               memory: 100Mi
             limits:
-              cpu: 800m
+              cpu: 4000m
               memory: 700Mi
           livenessProbe:
             exec:


rfranzke:
Sure, but why don't you create the image yourself locally? `make docker-images` should do the trick. :)

charts/internal/calico/templates/node/daemonset-calico-node.yaml
```diff
@@ -178,7 +178,7 @@ spec:
               cpu: 250m
               memory: 100Mi
             limits:
-              cpu: 800m
+              cpu: 4000m
```
ialidzhikov: With gardener/gardener-extension-networking-calico#14 we raised the limit to 800m and the calico-node Pods which were cpu throttled at `500m` are now with cpu usage of 50m.
At least it looks to me like the high cpu usage of calico-node is a result of another issue.

ialidzhikov:
> Sure, but why don't you create the image yourself locally? make docker-images should do the trick. :)

Yes, 👍

charts/internal/calico/templates/node/daemonset-calico-node.yaml
majst01: The hard part is to figure out the root cause, @zanetworker already helped a lot finding this issue. But is difficult.

We build the image, no prob

zanetworker:
@rfranzke and @ialidzhikov do you think it makes sense if we expose these values in the config?

ialidzhikov:
> @rfranzke and @ialidzhikov do you think it makes sense if we expose these values in the config? The problem is that every environment has different needs.

It make sense as there is no autoscaler for calico-node.

rfranzke:
@ialidzhikov there is calico-vpa, or? Shouldn't it help?

zanetworker:
@ialidzhikov we can have one like typha but it wouldn't help because the problem is not about how many nodes are in the cluster, rather how many pods there is per node. Also the cause of the spike in @majst01's case is not yet clear, but in anyways, I think it would help in such cases.

ialidzhikov:
> @ialidzhikov there is calico-vpa, or? Shouldn't it help?

Currently we deploy calico-typha-horizontal-autoscaler and calico-typha-vertical-autoscaler but they are both for calico-typha-deploy needs, if you mean this.

zanetworker:
@rfranzke we can have proportional autoscaler for calico-node as well.

majst01:
Running our build with 4000m limit:
```
k --kubeconfig seed top pod -n kube-system
NAME                                                              CPU(cores)   MEMORY(bytes)   
addons-nginx-ingress-controller-5b64594c9f-4rdxb                  23m          225Mi           
addons-nginx-ingress-nginx-ingress-k8s-backend-79cc57668c-fnz25   1m           59Mi            
blackbox-exporter-5d7968d895-bxx29                                1m           24Mi            
calico-kube-controllers-d574b6c9c-hdv42                           4m           25Mi            
calico-node-t2dld                                                 502m         61Mi            
calico-node-wsmkn                                                 4004m        57Mi            
calico-node-zvzck                                                 2718m        58Mi            
coredns-f8b6b6d6d-9gvq9                                           16m          27Mi            
coredns-f8b6b6d6d-fzp69                                           11m          28Mi            
kube-proxy-42bzs                                                  2m           79Mi            
kube-proxy-8z6nc                                                  61m          82Mi            
kube-proxy-qwlgt                                                  39m          84Mi            
metrics-server-564768c986-gb6jh                                   11m          55Mi            
node-exporter-76kj6                                               9m           30Mi            
node-exporter-lfmr8                                               13m          27Mi            
node-exporter-qk2fx                                               1m           29Mi            
node-problem-detector-fgl9f                                       200m         45Mi            
node-problem-detector-s7r7b                                       2m           63Mi            
node-problem-detector-st4t2                                       1m           43Mi            
vpn-shoot-54d7654888-62clw                                        5m           11Mi
```
something is weird, will dive into next week

zanetworker:
@majst01 I agree that it needs more debugging (would be interested in the cause myself of the conntrack explosion). In the mean time, I will add an autoscaler for calico-node.

zanetworker:
@majst01 I think #22 should address the comments above.

majst01:
Raised an issue in the calico-node repo the get help from the calico people eventually:

https://github.com/projectcalico/node/issues/447

zanetworker:
@stoyanr closing in favor of  #22 please re-open if you still think this is needed.

PR #24, opened 2020-04-11
Update calico-cpva to v0.8.2

### LINKED ISSUE #8: calico-typha-vertical-autoscaler in CrashLoopBackOff when deployments is available under multiple API groups
**What happened**:
calico-typha-vertical-autoscaler fails when deployments is available under multiple API groups.

**How to reproduce it (as minimally and precisely as possible)**:
1. Install linkerd. See https://linkerd.io/2/getting-started/

2. Ensure that there are multiple API groups serving resource `deployments`:

```
$ k api-resources
NAME                              SHORTNAMES   APIGROUP                       NAMESPACED   KIND
# ...
daemonsets                        ds           apps                           true         DaemonSet
deployments                       deploy       apps                           true         Deployment
# ...
daemonsets                        ds           tap.linkerd.io                 true         Tap
deployments                       deploy       tap.linkerd.io                 true         Tap
```

3. Ensure that `calico-typha-vertical-autoscaler` fails with

```
$ k logs calico-typha-vertical-autoscaler-7b9954b9df-ptrpq -n kube-system
I0217 20:41:29.699612       1 autoscaler.go:46] Scaling namespace: kube-system, target: deployment/calico-typha-deploy
E0217 20:41:30.799782       1 autoscaler.go:49] unknown target kind: Tap
```

**Anything else we need to know?**:

**Environment**:

- Gardener version (if relevant):
- Extension version: v1.3.0
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- Others:

ialidzhikov:
Update calico-cpva to v0.8.2.

Fixes #8 

**Release note**:

```improvement operator
An issue causing cpva to fail starting when there are multiple API resources with duplicating names but different kinds is now fixed.
```

### DIFF
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -25,5 +25,5 @@ images:
   tag: 1.7.1
 - name: calico-cpva
   sourceRepository: github.com/kubernetes-incubator/cluster-proportional-vertical-autoscaler
-  repository: k8s.gcr.io/cpvpa-amd64
-  tag: v0.8.1
+  repository: eu.gcr.io/k8s-artifacts-prod/cpa/cpvpa-amd64
+  tag: v0.8.2


charts/images.yaml
```diff
@@ -25,5 +25,5 @@ images:
   tag: 1.7.1
 - name: calico-cpva
   sourceRepository: github.com/kubernetes-incubator/cluster-proportional-vertical-autoscaler
-  repository: k8s.gcr.io/cpvpa-amd64
-  tag: v0.8.1
\ No newline at end of file
+  repository: eu.gcr.io/k8s-artifacts-prod/cpa/cpvpa-amd64
```
rfranzke, line 28: FYI - https://groups.google.com/forum/#!msg/kubernetes-dev/MkXnkTSJ_vs/IeI3LgmDCAAJ

--> Is it valid to use `k8s-artifacts-prod` GCR already?

ialidzhikov: See https://github.com/kubernetes-incubator/cluster-proportional-vertical-autoscaler/issues/33#issuecomment-612258918.

zanetworker: is this a temporary domain for the registry, i.e., will the image always be accessible at that place or will we have to change that later? The thread is not clear about this point.

ialidzhikov: @MrHohn, @listx, we can proceed with the `eu.gcr.io/k8s-artifacts-prod/cpa/cpvpa-amd64:v0.8.2` image, right?
When the flip for the vanity domain happens, we can switch back to `k8s.gcr.io/cpa/cpvpa-arm64:v0.8.2`, right?

listx: @ialidzhikov That's correct.

@zanetworker The `k8s.gcr.io` domain will probably be around a very, very long time. We've gone to very great lengths to try to keep it alive, and will continue to do so for the foreseeable future.

zanetworker: thanks for the swift response :)

zanetworker: I guess this here is also the same discussion https://github.com/kubernetes/kubernetes/pull/90093#discussion_r407580184

listx: Yes [kubernetes/kubernetes#90093 (comment)](https://github.com/kubernetes/kubernetes/pull/90093#discussion_r407580184) seems to be the same; I commented there also.

> thanks for the swift response :)

NP

PR #34, opened 2020-06-06
support image override at shoot level for calico images

mandelsoft:

/area control-plane
/kind enhancement
/priority normal
/area networking

This PR extends the provider config for the calico networking are to support
override of dedicated images for pods deployed by the extension.

This could simplify/enable testing new images in the existing dev landscape
without hampering the all the other shoots, because the usage of the images in question can be
restricted to a dedicated shoot.

Fixes #

**Release note**:

```noteworthy operator
If a shoot has problems with calico networking, check whether the shoot is still using
the standard image by checking the `networking/providerConfig/images` section
in the shoot manifest.
```


rfranzke:
This would allow end-users to overwrite the images used for Calico also for production clusters. Probably we don't want to allow this, right?
Instead, for example the image vector functionality could be extended to support such use-cases. Adding it to the API "just for development purposes" seems wrong to me.

vpnachev:
In the dev landscapes usually reconciliation can be simply disabled - this gives you the ability to change whatever system component you want. 
Also, a local landscape where you have full control without affecting anybody else should be used for such scenarios.

Generally - exposing this config to the end user could break down the whole cluster. And from security point of view - gardener operators cannot ensure that maintained and non-vulnerable images are used.

/reviewed do-not-merge
/area security
/kind api-change
/status author-action

zanetworker:
@mandelsoft do you still want to keep this PR?

mandelsoft:
To be honest, I don't know how I should test the shoot creation and wake-up process with a new networking without changing the networking for all clusters. And this is something I definitely don't want to do.

We can already override so many things, including the complete API server setup, so what is the problem with the networking. Sure, if you do something wrong it fails. But there are already so many places where I can destroy everything.

vpnachev:
>To be honest, I don't know how I should test the shoot creation and wake-up process with a new networking without changing the networking for all clusters. And this is something I definitely don't want to do.

It is very simple - create your own gardener installation where you have only single shoot cluster to test your development changes.


> We can already override so many things, including the complete API server setup, so what is the problem with the networking. Sure, if you do something wrong it fails. But there are already so many places where I can destroy everything.

Could you name a few where the end user is allowed to bring own container images?
If you really like to overwrite what Gardener deploys in the shoot, then the better approach is to go with mutating webhooks, but not to expose configurations via the API that can never be validated.

zanetworker:
@mandelsoft what is the status of this PR?

zanetworker:
closing for now, @mandelsoft if you still think this is a needed feature and want to discuss further you can re-open.

PR #37, opened 2020-06-22
Restore now calls Reconcile instead of being an empty op

plkokanov:

/area control-plane-migration
/kind task
/priority normal

Makes the Restore operation call Reconcile instead of just returning nil.
Also made the Migrate() operation only return nil as otherwise network managed resources can be deleted. Currently the logic to clean up all managed resources without deleting the corresponding objects from the shoot is part of the gardenlet migration flow.

Part of [#1631](https://github.com/gardener/gardener/issues/1631)

**Release note**:

```improvement operator
Restore now calls reconcile instead of just returning nil.
```


plkokanov:
Same as the PR for cilium.

PR #39, opened 2020-06-23
[gardener/gardener#2462] Add nodeSelector to run only on allowed worker pools

jannickfahlbusch:
/priority normal
/area networking

This makes the the following components run only on pools that are allowed to host system components and extensions:
* calico-kube-controller
* calico-node-vertical-autoscaler
* calico-typha-deploy
* calico-typha-horizontal-autoscaler
* calico-typha-vertical-autoscaler

Relates to https://github.com/gardener/gardener/issues/2462
Needs https://github.com/gardener/gardener/pull/2480

**Release note**:

```improvement operator
All non-DaemonSet resources now respect the node label `worker.gardener.cloud/system-components=true` as introduced by https://github.com/gardener/gardener/pull/2480
```


jannickfahlbusch:
/hold

This can only be merged after https://github.com/gardener/gardener/pull/2480 got merged

zanetworker:
> Hm, if we merge it like this then the Calico extension would not be compatible with older Gardener versions anymore. Probably it's safer to let the reconciler check the respective `Shoot`, and only add the node selector in case it detects that it has worker pool that explicitly allows system components (this way it knows that the shoot was reconciled by a new Gardener version).

Was about to say that this won't work if the workers do not have that label.

jannickfahlbusch:
This would need to be handled here right?

https://github.com/gardener/gardener-extension-networking-calico/blob/d7c117e2b59d31dcd5ac57684eab718e9d52d53b/pkg/charts/utils.go#L114-L139

zanetworker:
But since you would need to fetch the value from the shoot, you need access to the cluster object: 
https://github.com/gardener/gardener-extension-networking-calico/blob/d7c117e2b59d31dcd5ac57684eab718e9d52d53b/pkg/controller/actuator_reconcile.go#L53 

Then pass it to compute the chart values.

zanetworker:
/lgtm

@rfranzke since you reviewed this initially, do you want to have another look?

jannickfahlbusch:
We need to wait for the next release of gardener/gardener to vendor the changes here right?

vpnachev:
You can vendor even the master, but a new minor release is planned to be released soon, so waiting for it is also an option.

zanetworker:
@jannickfahlbusch Okay I see you are making use of `SystemComponentsAllowed`, so yes but functionality wise there is no dependencies.

charts/internal/calico/templates/typha/deployment-calico-typha.yaml
```diff
@@ -49,6 +49,9 @@ spec:
     spec:
       nodeSelector:
         beta.kubernetes.io/os: linux
```
vpnachev, line 51: I see you have removed this selector for the `calico-kube-controller`, is it possible to remove it also for `calico-typha`?

charts/internal/calico/templates/kube-controller/deployment-calico-kube-controllers.yaml
```diff
@@ -34,7 +34,7 @@ spec:
         gardener.cloud/role: system-component
     spec:
       nodeSelector:
-        beta.kubernetes.io/os: linux
+{{ toYaml .Values.nodeSelector | indent 8 }}
```
vpnachev, line 37:
```suggestion
      {{- if .Values.nodeSelector }}
      nodeSelector:
{{ toYaml .Values.nodeSelector | indent 8 }}
      {{- end }}
```

You can omit the node selector if it is not set in the values.

charts/internal/calico/values.yaml
```diff
@@ -24,4 +24,6 @@ images:
   calico-kube-controllers: "image-repository:image-tag"
   calico-podtodaemon-flex: "image-repository:image-tag"
   calico-cpva: "image-repository:image-tag"
-  calico-cpa: "image-repository:image-tag"
\ No newline at end of file
+  calico-cpa: "image-repository:image-tag"
+
+nodeSelector: {}
```
vpnachev, line 29: If you decide to implement the above suggestion, you have to comment this line.
```suggestion
# nodeSelector: {}
```

charts/internal/calico/templates/typha/deployment-calico-typha.yaml
jannickfahlbusch: I've removed it by accident. I'll readd it

charts/internal/calico/values.yaml
jannickfahlbusch: Done

charts/internal/calico/templates/kube-controller/deployment-calico-kube-controllers.yaml
jannickfahlbusch: Done

charts/internal/calico/templates/typha/deployment-calico-typha.yaml
jannickfahlbusch: Removed the selector

jannickfahlbusch:
@vpnachev @zanetworker I've worked in the changes.

Please take a look again

pkg/charts/utils.go
```diff
@@ -135,6 +136,13 @@ func ComputeCalicoChartValues(network *extensionsv1alpha1.Network, config *calic
 		},
 		"config": calicoConfig,
 	}
+
+	if workerSystemComponentsActivated {
+		calicoChartValues["nodeSelector"] = map[string]string{
+			gardenv1beta1constants.LabelWorkerPoolSystemComponents: "True",
```
rfranzke, line 142: This should be `true`

pkg/charts/charts_test.go
```diff
@@ -353,9 +353,57 @@ var _ = Describe("Chart package test", func() {
...
+						},
+					},
+					"ipv4": map[string]interface{}{
+						"pool":                string(poolIPIP),
+						"mode":                string(always),
+						"autoDetectionMethod": nil,
+					},
+				},
+				"nodeSelector": map[string]string{
+					"worker.gardener.cloud/system-components": "True",
```
rfranzke, line 398: `true`

PR #45, opened 2020-08-11
Expose monitoring configuration for Calico

zanetworker:

/kind enhancement
/priority normal
/area monitoring
/area networking

This PR exposes metrics and dashboards for Felix and Typha. 

Fixes #

**Release note**:

```improvement operator
Monitoring is now enabled by default which would expose metrics/dashboards for the typha and felix components. 
```


charts/internal/calico-monitoring/charts/utils-templates/Chart.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
```
rfranzke, line 1: This whole chart could be symlinked instead of copied

charts/internal/calico/templates/node/daemonset-calico-node.yaml
```diff
@@ -101,11 +101,23 @@ spec:
         # host.
         - name: calico-node
           image: {{ index .Values.images "calico-node" }}
+          ports:
+            {{- if .Values.config.monitoring.enabled }}
```
rfranzke, line 105: Out of curiosity: What is the rationale of making this monitoring configurable? When would I want to disable it?

example/20-network.yaml
```diff
@@ -50,6 +50,8 @@ spec:
     ipam:
       type: host-local
       cidr: usePodCIDR
+#    monitoring:
+#      enabled: true
```
rfranzke, line 54:
```suggestion
#   monitoring:
#     enabled: true
```

pkg/charts/utils.go
```diff
@@ -220,5 +234,9 @@ func generateChartValues(config *calicov1alpha1.NetworkConfig) (*calicoConfig, e
 		c.VethMTU = *config.VethMTU
 	}
 
+	if config.Monitoring != nil {
+		fmt.Printf("Monitoring enable? %t", config.Monitoring.Enabled)
```
rfranzke, line 238: left over?

charts/internal/calico/templates/node/daemonset-calico-node.yaml
zanetworker: if it causes unnecessary overhead on the seed Prometheus is one case, also for test clusters.

rfranzke: > if it causes unnecessary overhead on the seed Prometheus is one case, also for test clusters.

Hm, ok, I'm asking because this flag is exposed to end-users. Hence, when - as an end-user - would I disable this? Probably never as "I don't know/care" about the Prometheus running in the seed, so I just always enable it.
If there are no other use-cases, wouldn't it make more sense to not expose it in the `NetworkConfig` but do it as you suggested - not enable it for `testing` purpose cluster, for example?

> One other thing, if we wish to change the ports in the future for whatever reason

Didn't get this one, can't we change them either way?

zanetworker: > Didn't get this one, can't we change them either way?

We'd have to create a new release for that since we will be modifying the charts or? 

I see your point. Exposing it to the users does not bring any meaning, however, I am not sure if we want to tweak anything in the future related to monitoring, we currently don't have the distinction between an end-user or an operator when the shoot is modified. But if we do, then we can hide fields like this one, and expose it to operators only. 

That said, I have no strong opinion regarding disabling it.

rfranzke: Hm, sounds theoretical to me. We don't have these needs today, and there is no additional configuration option yet for these monitoring things --> I wouldn't even bother to introduce this in the API and save same code, and rather revisit when it gets relevant again (if we don't have some use-cases today).
I kind of get the point with the `testing` clusters, although such clusters don't get a monitoring stack in the first place, so these configuration artefacts cannot put any load on the non-existing Prometheus.

To sum up, I would save this additional complexity+code (yes, it's not very complex, but still) for now, but that's just my high level recommendation without being into the details too much.

charts/internal/calico-monitoring/calico-typha-dashboard.json
```diff
@@ -0,0 +1,2360 @@
...
+  "templating": {
+    "list": []
+  },
+  "time": {
+    "from": "now-6h",
+    "to": "now"
+  },
+  "timepicker": {},
+  "timezone": "",
+  "title": "Typha Dashborad (Calico)",
```
wyb1, line 2354: typo

charts/internal/calico-monitoring/calico-felix-dashboard.json
```diff
@@ -0,0 +1,1480 @@
...
+      "datasource": null,
+      "gridPos": {
+        "h": 1,
+        "w": 24,
+        "x": 0,
+        "y": 17
+      },
+      "id": 35,
+      "panels": [],
+      "title": "Route table",
```
wyb1, line 1168: Could you add a description for these panels?

charts/internal/calico-monitoring/calico-typha-dashboard.json
```diff
@@ -0,0 +1,2360 @@
+{
```
wyb1, line 1: None of the panels in this dashboard have descriptions. Can you add descriptions to the panels?

charts/internal/calico-monitoring/templates/calico-monitoring-config.yaml
```diff
@@ -0,0 +1,60 @@
...
+      - source_labels: [__meta_kubernetes_service_name]
+        regex: felix-metrics-svc
+        action: keep
+      - target_label: __address__
+        replacement: kube-apiserver:443
+      - source_labels: [__meta_kubernetes_pod_name]
+        regex: (.+)
+        target_label: __metrics_path__
+        replacement: /api/v1/namespaces/kube-system/pods/${1}:9091/proxy/metrics
+    - job_name: 'typha_metrics'
```
wyb1, line 32: same as above

charts/internal/calico/templates/node/felix-monitoring-service.yaml
```diff
@@ -0,0 +1,12 @@
+{{- if .Values.config.monitoring.enabled }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: felix-metrics-svc
```
vpnachev, line 5: Please, align the file name with the resource name. 
Either 
```suggestion
  name: felix-monitoring
```
Or rename the file to `felix-metrics-service.yaml`.

charts/internal/calico/templates/typha/typha-monitoring-service.yaml
```diff
@@ -0,0 +1,12 @@
+{{- if .Values.config.monitoring.enabled }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: typha-metrics-svc
```
vpnachev, line 5: Also align resource and file names.

pkg/controller/actuator.go
```diff
@@ -26,6 +27,8 @@ import (
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	gardenerkubernetes "github.com/gardener/gardener/pkg/client/kubernetes"
```
vpnachev, line 31: Move this import in the group of other `github.com/gardener/*` imports.

charts/internal/calico/templates/node/felix-monitoring-service.yaml
```diff
@@ -0,0 +1,13 @@
+{{- if .Values.config.monitoring.enabled }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: felix-monitoring
```
rfranzke, line 5:
```suggestion
  name: calico-felix-monitoring
```

charts/internal/calico/templates/typha/typha-monitoring-service.yaml
```diff
@@ -0,0 +1,13 @@
+{{- if .Values.config.monitoring.enabled }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: typha-monitoring
```
rfranzke, line 5:
```suggestion
  name: calico-typha-monitoring
```

PR #52, opened 2020-09-30
Typha and kube-controller have proper tolerations

mvladev:

/kind bug
/priority normal
/area networking

Those components are now always scheduled to ensure that the node network can work.

n/a

**Release note**:

```improvement operator
Calico Typha and kube-controller now tolerate `NoSchedule` and `NoExecute` taints to ensure that networking is always available.
```

### DIFF
--- a/charts/internal/calico/templates/kube-controller/deployment-calico-kube-controllers.yaml
+++ b/charts/internal/calico/templates/kube-controller/deployment-calico-kube-controllers.yaml
@@ -38,11 +38,14 @@ spec:
 {{ toYaml .Values.nodeSelector | indent 8 }}
       {{- end }}
       tolerations:
-        # Mark the pod as a critical add-on for rescheduling.
-        - key: CriticalAddonsOnly
-          operator: Exists
-        - key: node-role.kubernetes.io/master
-          effect: NoSchedule
+      # Make sure kube-controllers gets scheduled on all nodes.
+      - effect: NoSchedule
+        operator: Exists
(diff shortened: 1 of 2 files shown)


vpnachev (approved):
/lgtm

rfranzke (approved):
/lgtm

dguendisch:
We recently had the case were a "suspicious" node was cordoned/drained by a customers or operator, unfortuantely calico-kube-controllers (after the pod being manually deleted) was a few times rescheduled to the cordoned node (because it obviously tolerates the `NoSchedule` taint).
I'm wondering whether this toleration really makes sense, i.e. doesn't it achieve more the opposite? We often preserve suspicious nodes for later/offline analysis by simply draining them (and thus assuming that nothing relevant gets scheduled onto them anymore). In such cases tolerating `NoSchedule` has a negative effect.
So the only scenario I can imagine for a controller to tolerate `NoSchedule` is when there are _only_ nodes marked with `NoSchedule` in the whole cluster, but then again also no CoreDNS would run (as it doesn't tolerate NoSchedule).
@vpnachev @rfranzke do you recall whether you had other scenarios in mind with the changes from this PR?
cc @ScheererJ @DockToFuture

vpnachev:
> @vpnachev @rfranzke do you recall whether you had other scenarios in mind with the changes from this PR?

No. Just not sure if the nodes are created with this taint the networking cannot start-up, they will never got to healthy state, but if  this was the case, it should be easily verified by simply trying to create such cluster and its nodes should never get into ready state.

PR #55, opened 2020-10-02
Disable cache for Secrets and ConfigMaps

timebertt:

/area cost
/kind enhancement
/priority normal
/area networking

This PR disables the cache for Secrets and ConfigMaps, as there might be lots of them in large Seed clusters and the extension controller does barely use them and thus doesn't need to cache them.
It uses Secrets for applying ManagedResources and ConfigMaps for the monitoring config.

Fixes #

✅ Depends on gardener/gardener#2940

I tested this change on a medium-sized landscape and the memory footprint dropped from ~360Mi to ~160Mi

**Release note**:

```improvement operator
The cache for Secrets and ConfigMaps has been disabled to decrease the extension controller's memory footprint.
```


cmd/gardener-extension-networking-calico/app/app.go
```diff
@@ -75,7 +77,13 @@ func NewControllerManagerCommand(ctx context.Context) *cobra.Command {
...
 			util.ApplyClientConnectionConfigurationToRESTConfig(configFileOpts.Completed().Config.ClientConnection, restOpts.Completed().Config)
 
-			mgr, err := manager.New(restOpts.Completed().Config, mgrOpts.Completed().Options())
+			completedMgrOpts := mgrOpts.Completed().Options()
+			completedMgrOpts.NewClient = utils.NewClientFuncWithDisabledCacheFor(
+				&corev1.Secret{},    // applied for ManagedResources
+				&corev1.ConfigMap{}, // applied for monitoring config
+			)
+
+			mgr, err := manager.New(restOpts.Completed().Config, completedMgrOpts)
```
timebertt, line 86: @timuthy WDYT about the usage shown here?
Do you think it makes sense, to make this somehow configurable in `controllercmd.ManagerOptions`?

timuthy: For now it is sufficient and simple, no need to make it configurable for operators, imo. We can still do it if it turns out to be necessary.

DockToFuture (dismissed):
/lgtm

timebertt:
Rebased and vendored gardener/gardener@985629a24572 including (gardener/gardener#2940).
/ready

timuthy (approved):
/lgtm

DockToFuture (approved):
/lgtm

PR #59, opened 2020-10-28
Update calico to v3.16.4

DockToFuture:

/kind enhancement
/priority normal
/area networking

Update calico to v3.16.4 for cluster >= `1.16` kubernetesVersion else use calico `v3.13.5`

Would be nice if someone else could test the PR as well 

**Release note**:

```improvement operator
Update calico version to `v3.16.4` for cluster >= `1.16` kubernetesVersion
```


charts/internal/calico/templates/custom-resource-definition/crd-bgpconfigurations.yaml
```diff
@@ -1,15 +1,135 @@
 ---
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
rfranzke, line 2: This means that we cannot install calico on clusters < 1.16 anymore. The `apiextensions.k8s.io/v1` API group is only available in 1.16+.
Probably this calico version is not compatible with Kubernetes < 1.16, or? Can you check this? If yes we probably have to stay with 3.13 for Kubernetes < 1.16, and use the 3.16 only for Kubernetes >= 1.16.

charts/internal/calico/templates/custom-resource-definition/crd-bgpconfigurations.yaml
```diff
@@ -1,15 +1,135 @@
...
 kind: CustomResourceDefinition
 metadata:
   name: bgpconfigurations.crd.projectcalico.org
+  annotations:
+    controller-gen.kubebuilder.io/version: (devel)
```
rfranzke, line 7: Why do we need this?

rfranzke (changes requested):

charts/internal/calico/templates/custom-resource-definition/crd-bgpconfigurations.yaml
DockToFuture: That is a good point, will check that.

charts/internal/calico/templates/custom-resource-definition/crd-bgpconfigurations.yaml
```diff
@@ -13,3 +14,138 @@ spec:
     kind: BGPConfiguration
     plural: bgpconfigurations
     singular: bgpconfiguration
+{{else}}
```
rfranzke, line 17:
```suggestion
{{- else }}
```

(same for other occurrences)

charts/internal/calico/templates/custom-resource-definition/crd-bgpconfigurations.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if semverCompare "< 1.16" .Values.global.kubernetesVersion  }}
```
rfranzke, line 1: You can use 
```suggestion
{{- if semverCompare "< 1.16" .Capabilities.KubeVersion.GitVersion  }}
```

(same for other occurrences

pkg/charts/utils.go
```diff
@@ -147,7 +147,8 @@ func ComputeCalicoChartValues(network *extensionsv1alpha1.Network, config *calic
 			calico.ClusterProportionalVerticalAutoscalerImageName: imagevector.ClusterProportionalVerticalAutoscalerImage(),
 		},
 		"global": map[string]string{
-			"podCIDR": network.Spec.PodCIDR,
+			"podCIDR":           network.Spec.PodCIDR,
+			"kubernetesVersion": kubernetesVersion,
```
rfranzke, line 151: not needed

pkg/charts/utils.go
```diff
@@ -127,7 +127,7 @@ func (c *calicoConfig) toMap() (map[string]interface{}, error) {
 }
 
 // ComputeCalicoChartValues computes the values for the calico chart.
-func ComputeCalicoChartValues(network *extensionsv1alpha1.Network, config *calicov1alpha1.NetworkConfig, workerSystemComponentsActivated bool) (map[string]interface{}, error) {
+func ComputeCalicoChartValues(network *extensionsv1alpha1.Network, config *calicov1alpha1.NetworkConfig, workerSystemComponentsActivated bool, kubernetesVersion string) (map[string]interface{}, error) {
```
rfranzke, line 130:
```suggestion
func ComputeCalicoChartValues(network *extensionsv1alpha1.Network, config *calicov1alpha1.NetworkConfig, workerSystemComponentsActivated bool) (map[string]interface{}, error) {
```

charts/internal/calico/values.yaml
```diff
@@ -1,4 +1,6 @@
+
 global:
+  kubernetesVersion: 1.18.0
```
rfranzke, line 3: not needed

charts/internal/calico/values.yaml
```diff
@@ -1,4 +1,6 @@
+
```
rfranzke, line 1:
```suggestion
```

rfranzke (changes requested):

pkg/charts/utils.go
DockToFuture: Why would i remove here the `kubernetesVersion`? I need it for the image vector to resolve the image version?

rfranzke: You're right, I missed that. Ignore this suggestion, thanks.

rfranzke (approved):
/lgtm from the changes, but I haven't tested it and don't have time to do so

DockToFuture:
Tested it locally for an older and newer k8s version.
```
shoot-gcp-new   gcp            1.18.4    gcp    shoot-gcp-new.dev.dev.k8s-hana.ondemand.com   Awake         Succeeded   100        True        True      True    True     49m

shoot-gcp-old   gcp            1.15.12   gcp    shoot-gcp-old.dev.dev.k8s-hana.ondemand.com   Awake         Succeeded   100        True        True      True    True     49m
```

rfranzke:
@DockToFuture did you also test the update of an existing 1.15 cluster to 1.16?

DockToFuture:
Yes updated the cluster from `1.15.12` to  `1.16.10`. The crds and charts were updated successful.

rfranzke:
OK, thanks @DockToFuture. I'm okay with merging unless there is somebody else that might want to review (@mvladev?).
Please check and merge as it fits.

PR #64, opened 2020-12-01
Update calico to v3.17.1

DockToFuture:

/area networking
/kind enhancement
/priority normal

Update calico version to `v3.17.1` for cluster >= `1.16` kubernetesVersion.

**Release note**:

```improvement operator
Update calico version to `v3.17.1` for cluster >= `1.16` kubernetesVersion.
```


DockToFuture:
**Creation** was tested for:
aws:
-  [x] 1.19.2
-  [x] 1.15.12

az:
-  [x] 1.19.2
-  [x] 1.15.12

gcp:
- [x] 1.18.4
- [x] 1.15.12

**Migration** was tested for:
aws:
-  [x] 1.15.12 -> 1.16.12
 
az:
-  [] 1.15.12 -> 1.16.12

gcp:
- [] 1.15.12 -> 1.16.12

charts/images.yaml
```diff
@@ -1,8 +1,8 @@
 images:
 - name: calico-node
   sourceRepository: github.com/projectcalico/calico
-  repository: quay.io/calico/node
-  tag: v3.16.4
+  repository: eu.gcr.io/gardener-project/3rd/calico/node
```
rfranzke, line 4: Where are these images coming from / how are they published here?

DockToFuture: I uploaded them manually, they switched from `quay.io` to `docker`.

ialidzhikov: I believe this should be not a problem and we should continue using the upstream docker images.

DockToFuture: We run into rate limits if we use the upstream images.

ialidzhikov: I don't get, they should run with "pullPolicy: IfNotPresent" and this should not be a problem.

DockToFuture: I run even locally into rate limits, I don't care using them but I would personally feel better knowing that we don't run into that issue.

ialidzhikov: We did similar workaround for golang and alpine images as they are used to build the extension images because of an issue with the ci/cd that is always pulling images and is not able to cache the images AFAIK.

DockToFuture: Then I will adapt it.

ialidzhikov: Thank you!

DockToFuture: Upgraded the PR :)

rfranzke (approved):
/needs second-opinion

DockToFuture:
Would be cool if someone could give it a second check, would like to merge it and build a new release.

ialidzhikov:
> Would be cool if someone could give it a second check, would like to merge it and build a new release.

I will have a look.

PR #67, opened 2021-01-29
Vendor to gardener v1.16.0 and update controller runtime to v0.7.1

DockToFuture:

/area networking
/kind enhancement
/priority normal

Vendor to gardener v1.16.0 and update controller runtime to v0.7.1

**Release note**:

```other operator
Vendor to gardener v1.16.0 and update controller runtime to v0.7.1
```


Dockerfile
```diff
@@ -1,5 +1,5 @@
 ############# builder
-FROM golang:1.15.3 AS builder
+FROM golang:1.15.7 AS builder
```
vpnachev, line 2:
```suggestion
FROM eu.gcr.io/gardener-project/3rd/golang/golang:1.15.7 AS builder
```

Please, update also the alpine image. you can use `eu.gcr.io/gardener-project/3rd/alpine:3.12.3`.

vpnachev (commented):
One proposal regarding docker images.
Rest is looking good to me.

Dockerfile
DockToFuture: you probably mean `FROM eu.gcr.io/gardener-project/3rd/golang:1.15.7 AS builder` sure will update it.

vpnachev (approved):
/lgtm
/needs/second-opinion

rfranzke (approved):

PR #68, opened 2021-02-11
Add security context to install-cni container

DockToFuture:

/area networking
/kind enhancement
/priority normal

Add security context to install-cni container

**Release note**:

```other operator
Add security context to install-cni container
```

### DIFF
--- a/charts/internal/calico/templates/node/daemonset-calico-node.yaml
+++ b/charts/internal/calico/templates/node/daemonset-calico-node.yaml
@@ -96,6 +96,8 @@ spec:
             name: cni-bin-dir
           - mountPath: /host/etc/cni/net.d
             name: cni-net-dir
+        securityContext:
+          privileged: true
         # Adds a Flex Volume Driver that creates a per-pod Unix Domain Socket to allow Dikastes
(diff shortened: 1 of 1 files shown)


rfranzke:
Can you explain more why you propose this change? The change itself looks good, but it would be interesting to know the backgrounds.
/invite @mvladev

rfranzke:
@DockToFuture ?

DockToFuture:
Sometimes we are seeing that calico cni plugin is missing on the nodes and so far the logs say that it were successfully installed but it is missing. It was either not installed or somehow deleted afterwards. Since I wasn't able in reproducing this problem locally and the only difference to the provided charts from calico is, that we are lacking the security context in the init container. I would like to add the security context to see that it is not related and since we are using it anyhow in the calico node, I see no objections against it. If you however can exclude that, we can close the PR. :)

rfranzke (approved):
> and the only difference to the provided charts from calico is, that we are lacking the security context in the init container which install the binary.

Ah, I see, yeah, of course, we should follow what the Project Calico maintainers are providing as examples. Thanks for the explanation!

/lgtm

PR #72, opened 2021-03-01
Configure calico to use autodetection method cidr

DockToFuture:

/area networking
/kind enhancement
/priority normal

Configure calico to use `IP_AUTODETECTION_METHOD=cidr=cidrRange`

Calico uses by default (`IP_AUTODETECTION_METHOD=first-found`) the first found interface on the node, when there are custom interfaces it is possible that they are detected before the actual interface and thus resulting in a broken cluster.
Therefore is `IP_AUTODETECTION_METHOD` switched to `cidr`  mode with the corresponding node cidr from the shoot object to prevent the usage of the wrong interfaces. 

/cc: @ScheererJ @mandelsoft 

**Release note**:

```other operator
Calico is configured to use `IP_AUTODETECTION_METHOD=cidr=cidrRange` with the nodeCidr as cidrRange to prevent the using the wrong interface on the node.
```

### DIFF
--- a/pkg/controller/actuator_reconcile.go
+++ b/pkg/controller/actuator_reconcile.go
@@ -16,6 +16,7 @@ package controller
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/pkg/errors"
 
@@ -100,6 +101,14 @@ func (a *actuator) Reconcile(ctx context.Context, network *extensionsv1alpha1.Ne
 		if err != nil {
 			return err
 		}
+	}
+
+	if cluster.Shoot.Spec.Networking.Nodes != nil && len(*cluster.Shoot.Spec.Networking.Nodes) > 0 {
+		autodetectionMode := fmt.Sprintf("cidr=%s", *cluster.Shoot.Spec.Networking.Nodes)
+		if networkConfig == nil {
+			networkConfig = &calicov1alpha1.NetworkConfig{IPv4: &calicov1alpha1.IPv4{}}
+		}
+		networkConfig.IPv4.AutoDetectionMethod = &autodetectionMode
 	}
 
 	// Create shoot chart renderer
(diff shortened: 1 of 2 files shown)


pkg/controller/actuator_reconcile.go
```diff
@@ -102,6 +103,14 @@ func (a *actuator) Reconcile(ctx context.Context, network *extensionsv1alpha1.Ne
 		}
 	}
 
+	if cluster.Shoot.Spec.Networking.Nodes != nil && len(*cluster.Shoot.Spec.Networking.Nodes) > 0 {
+		autodetectionMode := fmt.Sprintf("cidr=%s", *cluster.Shoot.Spec.Networking.Nodes)
+		if networkConfig == nil {
+			networkConfig = &calicov1alpha1.NetworkConfig{IPv4: &calicov1alpha1.IPv4{}}
+		}
+		networkConfig.IPv4.AutoDetectionMethod = &autodetectionMode
```
ScheererJ, line 111: Would the following not be more concise:
networkConfig.IPv4.AutoDetectionMethod = "cidr=" + *cluster.Shoot.Spec.Networking.Nodes

If yes, we can get rid of line 19 and 107.

ScheererJ (approved):

pkg/controller/actuator_reconcile.go
DockToFuture: we need a pointer here right

ScheererJ: Ok, but the string concatenation should be possible without fmt, right?

DockToFuture: Sure, thats possible. Over the extensions we used `Sprintf` function but I have no strong opinion here.

rfranzke (approved):

PR #74, opened 2021-03-16
do not disable kubeController with backend none

mwennrich:
/area networking
/kind TODO
/priority normal

Currently setting `backend` to `none` in the shoot.spec disables the calico-kube-controllers: https://github.com/gardener/gardener-extension-networking-calico/blob/master/pkg/charts/utils.go#L178-L182

This even overrules the explicit setting of `kubeControllers.enabled: true`

This PR removes this behavior and therefore allows shoots with `backend: none` to have calico-kube-controllers installed.

Users of gardener-extension-networking-calico with `backend: none` who do not want kubeControllers enabled, should disable kubeControllers (`kubeControllers.enabled: false`) in the shoot spec.

Attention: this changes the default behavior for clusters with `backend: none`. Cluster-admins who wish to have kubeControllers disabled should make sure, that kubeControllers.enabled is set to false before upgrading. (Clusters with any other backend (bird, vxlan) are unaffected.)

**Release note**:
Allow enabling of calico-kube-controllers for clusters with `backend: none`
**Attention**: this changes the default behavior for clusters with `backend: none`. Cluster-admins who wish to have kubeControllers disabled should make sure, that kubeControllers.enabled is set to false before upgrading. (Clusters with any other backend (bird, vxlan) are unaffected.)

```other operator

```

### DIFF
--- a/pkg/charts/charts_test.go
+++ b/pkg/charts/charts_test.go
@@ -209,7 +209,7 @@ var _ = Describe("Chart package test", func() {
 						"enabled": trueVar,
 					},
 					"kubeControllers": map[string]interface{}{
-						"enabled": falseVar,
+						"enabled": trueVar,
 					},
 					"veth_mtu": defaultMtu,
 					"monitoring": map[string]interface{}{
--- a/pkg/charts/utils.go
+++ b/pkg/charts/utils.go
@@ -176,7 +176,6 @@ func generateChartValues(config *calicov1alpha1.NetworkConfig) (*calicoConfig, e
 		}
 	}
 	if c.Backend == calicov1alpha1.None {
-		c.KubeControllers.Enabled = false
 		c.Felix.IPInIP.Enabled = false
 		c.IPv4.Mode = calicov1alpha1.Never
 	}


gardener-robot-ci-3:
Thank you @mwennrich for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

danielfoehrKn:
@DockToFuture could you please check this PR when you got the time?  Thanks.

ScheererJ:
Hi @mwennrich,

could you please comment why you think that the kube-controllers are required in the calico scenario with backend set to none? In our use case, i.e. using the kubernetes datastore as backend for calico, only the node controller is relevant at all (see https://docs.projectcalico.org/reference/kube-controllers/configuration for details). However, we do not use any of the node specific calico resources which would need to be cleaned up, e.g. ipam blocks, when we do not use a backend.

Please share your scenario details.

Best regards

mwennrich:
Yes, sorry, you're right ... my initial thought was, that we need the policy controller for "policy-only" mode ... but most of these controllers are only needed when using etcd as datastore. Closing.

ScheererJ:
Indeed, the name calico-kube-controllers suggests a lot more than it actually does when used with kubernetes as datastore.
Thanks anyway for your pull request.

PR #85, opened 2021-05-25
Added usage of secondary-cni-dir to workaround random cni installation issues

ScheererJ:

/area networking
/kind enhancement

We have seen some rare cases where the cni installation seem to succeed, but the calico cni binaries are missing afterwards. This renders the node unusable as no new containers can be created. The workaround is to delete the calico-node pod to trigger the cni installation again.
This pull requests mounts the cni binary directory /opt/cni/bin a second time inside the install-cni container. The cni install container now effectively copies the cni binaries twice hopefully increasing the chance that the binaries will actually get copied or providing more insights into the issue.

This is kind of a workaround for the calico cni installation issues.

Tested with kubernetes 1.20.6 and 1.15.12.

**Release note**:

```other operator
None
```

### DIFF
--- a/charts/internal/calico/templates/node/daemonset-calico-node.yaml
+++ b/charts/internal/calico/templates/node/daemonset-calico-node.yaml
@@ -103,6 +103,8 @@ spec:
             name: cni-bin-dir
           - mountPath: /host/etc/cni/net.d
             name: cni-net-dir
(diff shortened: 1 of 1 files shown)


DockToFuture (approved):
/lgtm

PR #87, opened 2021-06-01
Update calico-cpa

DockToFuture:

/area networking
/kind enhancement

Update calico-cpa to verison 1.8.3. User ID has changed to `65532` as the `non-root` user is used.

**Release note**:

```other operator
Update calico-cpa to verison 1.8.3
```


ScheererJ (approved):
/lgtm

PR #91, opened 2021-06-21
Extend makefile for landscaper local dev

achimweigel:

/area dev-productivity
/kind enhancement

This PR add new makefile targets to create cnudie based component descriptors and landscaper installations for local dev setup.

Fixes #

**Release note**:

```other operator
NONE
```


Makefile
```diff
@@ -112,3 +115,26 @@ verify: check format test
...
 verify-extended: install-requirements check-generate check format test-cov test-cov-clean
+
+#####################################################################
+# Rules for cnudie component descriptors dev setup #
+#####################################################################
+
+.PHONY: cnudie-docker-images
+cnudie-docker-images:
+	@echo "Building docker images for version $(EFFECTIVE_VERSION) for registry $(CNUDIE_IMAGE_REGISTRY)"
+	@docker build -t $(CNUDIE_IMAGE_REGISTRY)/$(NAME):$(EFFECTIVE_VERSION) -f Dockerfile .
```
ScheererJ, line 126: Why do we have a separate docker build here instead of reusing the existing one and tagging the image as needed?
The command line option differ, e.g. with regards to memory limits. I do not like to duplicate code.

hack/get-cd-component-name.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+#
+# Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+#
+# SPDX-License-Identifier: Apache-2.0
+
+set -e
+
+echo "github.com/gardener/gardener-extension-networking-calico"
```
ScheererJ, line 9: Is it really necessary to have a bash script just for one configuration setting? Would it not be better to have multiple config options in one file together and source it?

hack/get-cd-registry.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+#
+# Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+#
+# SPDX-License-Identifier: Apache-2.0
+
+set -e
+
+echo "eu.gcr.io/sap-se-gcr-k8s-private/cnudie/gardener/development"
```
ScheererJ, line 9: Is it really necessary to have a bash script just for one configuration setting? Would it not be better to have multiple config options in one file together and source it?

hack/get-image-registry.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+#
+# Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
+#
+# SPDX-License-Identifier: Apache-2.0
+
+set -e
+
+echo "eu.gcr.io/gardener-project/gardener/extensions"
```
ScheererJ, line 9: Is it really necessary to have a bash script just for one configuration setting? Would it not be better to have multiple config options in one file together and source it?

hack/get-cd-registry.sh
ScheererJ, line 9: Does a reference to a private repository make sense in an open source project?

ScheererJ (changes requested):

hack/create-installation.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
```
vpnachev, line 1:
```suggestion
#!/usr/bin/env bash
```

hack/create-installation.sh
```diff
@@ -0,0 +1,56 @@
...
+#
+# Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
vpnachev, line 3:
```suggestion
# Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

hack/generate-cd.sh
```diff
@@ -0,0 +1,74 @@
+#!/bin/bash
```
vpnachev, line 1:
```suggestion
#!/usr/bin/env bash
```

hack/generate-cd.sh
```diff
@@ -0,0 +1,74 @@
...
+#
+# Copyright (c) 2020 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```
vpnachev, line 3:
```suggestion
# Copyright (c) 2021 SAP SE or an SAP affiliate company. All rights reserved. This file is licensed under the Apache Software License, v. 2 except as noted otherwise in the LICENSE file
```

hack/get-cd-component-name.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
vpnachev, line 1:
```suggestion
#!/usr/bin/env bash
```

hack/get-cd-registry.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
vpnachev, line 1:
```suggestion
#!/usr/bin/env bash
```

hack/get-image-registry.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
vpnachev, line 1:
```suggestion
#!/usr/bin/env bash
```

Makefile
achimweigel: I have changed this such that we have only one docker build target which is called with different settings.

hack/get-cd-component-name.sh
achimweigel: As proposed, now all configuration settings are in one file which is sourced.

hack/get-cd-registry.sh
achimweigel: As proposed, now all configuration settings are in one file which is sourced.

hack/get-image-registry.sh
achimweigel: As proposed, now all configuration settings are in one file which is sourced.

hack/get-cd-registry.sh
achimweigel: I have replaced the URL by a hint to insert the location for the component descriptor.

hack/generate-cd.sh
achimweigel: changes done

hack/get-cd-registry.sh
achimweigel: file removed

hack/get-image-registry.sh
achimweigel: file removed

hack/get-cd-component-name.sh
achimweigel: I habe removed the file

Makefile
```diff
@@ -112,3 +120,25 @@ verify: check format test
...
+	@if ! docker images $(CNUDIE_IMAGE_REGISTRY)/$(NAME) | awk '{ print $$2 }' | grep -q -F $(EFFECTIVE_VERSION); then echo "$(CNUDIE_IMAGE_REGISTRY)/$(NAME) version $(EFFECTIVE_VERSION) is not yet built. Please run 'make cnudie-docker-images'"; false; fi
+	@docker push $(CNUDIE_IMAGE_REGISTRY)/$(NAME):$(EFFECTIVE_VERSION)
+
+.PHONY: cnudie-cd-build-push
+cnudie-cd-build-push:
+	@EFFECTIVE_VERSION=$(EFFECTIVE_VERSION) ./hack/generate-cd.sh
+
+.PHONY: cnudie-create-installation
+cnudie-create-installation:
+	@EFFECTIVE_VERSION=$(EFFECTIVE_VERSION) ./hack/create-installation.sh
```
ScheererJ, line 144: Please add the missing newline at the end of the file.

hack/environment.sh
```diff
@@ -0,0 +1,27 @@
...
+  fi
+}
+
+get_cd_component_name () {
+  echo "github.com/gardener/gardener-extension-networking-calico"
+}
+
+get_image_registry () {
+  echo "eu.gcr.io/gardener-project/gardener/extensions"
+}
```
ScheererJ, line 27: Please add the missing newline at the end of the file.

hack/environment.sh
```diff
@@ -0,0 +1,27 @@
...
+    exit 1
+  fi
+}
+
+get_cd_component_name () {
+  echo "github.com/gardener/gardener-extension-networking-calico"
+}
+
+get_image_registry () {
+  echo "eu.gcr.io/gardener-project/gardener/extensions"
```
ScheererJ, line 26: I am a bit confused now. The usual image registry as defined in https://github.com/gardener/gardener-extension-networking-calico/blob/558fbe3ee1918e768f01d4a38d99c3d5d140f570/Makefile#L17 is
`eu.gcr.io/gardener-project/extensions`
while you use
`eu.gcr.io/gardener-project/gardener/extensions`.
Why is there an additional `gardener` in there? Can we simply reuse the existing registry?
In case we could use just one registry I would vote to unify the docker build and simply always give the image another tag, i.e. the resulting image always get VERSION, latest and VERSION-git-hash (=EFFECTIVE_VERSION).

What do you think?

ScheererJ (changes requested):

hack/environment.sh
achimweigel: I do not know the repository eu.gcr.io/gardener-project/gardener/extensions. This is not the one used when e.g. a release is made. Therefore this one is used eu.gcr.io/gardener-project/gardener/extensions. You find this repo also configured in .ci/pipeline_definitions. 

I do not know if we could unify the image repo as I do not know where eu.gcr.io/gardener-project/extensions is used. With respect to tagging I'm not sure if we should always add the tags like VERSION and latest. The goal of the new make targets is to create a consistent component descriptor with images for testing purposes which should not interfere with other use cases. Not sure if some other component relies on the version or latest tag.

Makefile
achimweigel: newline added

hack/environment.sh
ScheererJ: Thanks to your pull request we figured out that the tag used in the makefile was inconsistent to the one used in the continuous integration pipeline. Thanks for that.
The result is that we updated the makefile, which means you should rebase. Furthermore, now the base tag is always the same. The only thing different is the version.
I would vote for removing `CNUDIE_IMAGE_REGISTRY` from the makefile and simply always building with all tags, i.e. short version, effective version (including git has) and `latest`.

Makefile
```diff
@@ -23,6 +23,9 @@ LD_FLAGS                    := "-w -X github.com/gardener/$(EXTENSION_PREFIX)-$(
 LEADER_ELECTION             := false
 IGNORE_OPERATION_ANNOTATION := true
 
+EFFECTIVE_VERSION           := $(VERSION)-$(shell git rev-parse HEAD)
+CNUDIE_IMAGE_REGISTRY       := $(shell source "$(HACK_DIR)/environment.sh"; get_image_registry)
```
ScheererJ, line 27: Please remove this (see below).

ScheererJ (changes requested):

hack/environment.sh
achimweigel: I'm not sure if we should add short version and latest. The cnudie build targets are just for local Landscaper dev scenarios and we want to reduce the possibility to interfere with other use cases. Especially adding the latest tag might be quite risky, e.g. if you have a local buggy test commit not pushed to origin/master.

ScheererJ: I do not really get it. If this change is only to support the local landscaper development scenario why would it even matter if there are multiple container tags? You would only use the one with the effective version, i.e. git hash.
I really do not like duplicating things as I fear that it will diverge over time. Therefore, I vote for keeping just one docker build command.

@DockToFuture What do you think?

achimweigel: I think you are right. I thought that all tags will be added when pushed to the registry but this is not the case. I have removed the additional build target and added the effective version tag.

Makefile
achimweigel: I have removed CNUDIE_IMAGE_REGISTRY.

ScheererJ (approved):
/lgtm

PR #92, opened 2021-06-23
Enabled automatic management of the kubeconfig used by the cni operations

ScheererJ:

/area networking
/kind enhancement

Starting with kubernetes v1.21 service account tokens have a limited life time per default (BoundServiceAccountTokenVolume feature). Therefore, it is not sufficient to simply take the service account token during the init container of calico node and use it happily ever after. This change enables the automatic management by calico, i.e. if the service account token changes the cni kubeconfig will be updated automatically.

None.

**Release note**:

```other operator
The kubeconfig used by cni (/etc/cni/net.d/calico-kubeconfig) will be automatically updated if the service account token changes.
```


DockToFuture (approved):
/lgtm

rfranzke (approved):

PR #97, opened 2021-06-30
Removed support for very old kubernetes releases in terms of older priority class api versions.

ScheererJ:

/area networking
/kind technical-debt

Removed the helm template definition of the priority class api version. We only use it in one place and it is stable since quite some time.

None.

**Release note**:

```other operator

```


DockToFuture (approved):
/lgtm

DockToFuture (approved):
/lgtm

Issue #101, opened 2021-07-06
Remove bidirectional mount on /sys/fs


marwinski:
**How to categorize this issue?**

/area networking
/kind enhancement
/priority 1

**What would you like to be added**:

Please remove the bidirectional mount of `/sys/fs` in the calico-node container as this is currently not required.

**Why is this needed**:

With https://github.com/gardenlinux/gardenlinux/issues/275 we are investigating an issue where the `/sys/fs/cgroup` mount disappears from the node rendering the node unusable. As the kubelet still reports a healthy node pods are still scheduled on that node but never start. This situation can go on for days.

The `calico-node` container is currently the only container that we know that has been started and that can cause the issue with the bidirectional mount. This said, this might not be caused by a program in the calico-node container but we will only know this after this restriction has been applied.

PR #102, opened 2021-07-06
remove bidirectional mount propagation

DockToFuture:

/area networking
/kind enhancement

Remove bidirectional mount propagation from calico node.
The bidirectional mount propagation is not required and can lead to problems if a lazy unmount is done.

Fixes https://github.com/gardener/gardener-extension-networking-calico/issues/101

The PR was tested locally to check whether the bpf filesystem is properly mounted without the bidirectional mount propagation.

**Release note**:

```other operator
none
```


ScheererJ (approved):
/lgtm

PR #105, opened 2021-07-19
Disallow running without kube-proxy.

ScheererJ:
Calico only supports native service routing with the ebpf data plane, which is not yet supported by this extension.
Therefore, we reject the reconciliation of clusters with the kube-proxy flag to disabled.

/area networking
/kind enhancement

With https://github.com/gardener/gardener/pull/4260 it is possible to disable kube-proxy. Until we support the ebpf data plane of calico we disallow this for the calico networking extension by rejecting the cluster reconciliation.

None.

**Release note**:

```other operator

```


DockToFuture (approved):
/lgtm

PR #126, opened 2021-12-02
Increase memory limit for typha

DockToFuture:

/area networking
/kind enhancement

Increase memory limit of typha to `1000Mi` to better cope with load peaks when a lot of new node joins the cluster concurrently. 

**Release note**:

```other operator
Increase memory limit of typha to `1000Mi`.
```


ScheererJ (approved):
/lgtm

PR #131, opened 2021-12-07
fix autoscaling for node-cpva

DockToFuture:
Co-authored-by: Johannes Scheerer <johannes.scheerer@sap.com>

/area networking
/kind bug

Set `resources.gardener.cloud/preserve-resources=true` annotation so that GRM will respect the request values set by `node-cpva`.

**Release note**:

```other operator
Set `resources.gardener.cloud/preserve-resources=true` annotation so that GRM will respect the request values set by `node-cpva`.
```


ScheererJ (approved):
/lgtm

PR #145, opened 2022-01-26
Fixed vpa for calico-kube-controllers being applied even if the target was not deployed

ScheererJ:

/area networking
/kind bug

Fixed vpa for calico-kube-controllers being applied even if the target was not deployed.

In case no backend is used, calico-kube-controllers is not deployed. Therefore, the corresponding vpa resource should also not be deployed.

None.

**Release note**:

```other operator
Only apply vpa for calico-kube-controllers if calico-kube-controllers is actually deployed.
```


rfranzke (approved):
/lgtm

PR #147, opened 2022-02-07
Improve the calico scrape configuration

istvanballok:

/area monitoring
/kind enhancement

- Use server side filtering for the endpoint discovery based on the
  namespace
- Only scrape the port named "metrics", use the port_number label instead
  of hard coding the port number
- Keep the node and pod labels. Previously, without those labels the
  series overwrote each other and only one series was persisted in
  Prometheus.
- Only keep the felix and typha metrics. Under some circumstances,
  these scrape targets somehow exposed api server metrics which caused
  confusion.

Fixes #

**Release note**:

```other operator
Improve the calico (felix and typha) scrape configuration
```

### DIFF
--- a/charts/internal/calico-monitoring/templates/calico-monitoring-config.yaml
+++ b/charts/internal/calico-monitoring/templates/calico-monitoring-config.yaml
@@ -23,6 +23,8 @@ data:
       kubernetes_sd_configs:
       - role: endpoints
         api_server: https://kube-apiserver:443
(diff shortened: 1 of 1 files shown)


wyb1 (approved):
/lgtm

DockToFuture (approved):
/lgtm

ScheererJ (approved):
/lgtm

PR #149, opened 2022-02-08
Update calico typha cpva to not autoscale at the same step size for calico typha cpva and calico node cpva.

DockToFuture:

/area networking
/kind enhancement

Update calico typha cpva to not autoscale at the same step size for calico typha cpva and calico node cpva.

Fixes #

**Release note**:

```other operator
Update calico typha cpva to not autoscale at the same step size for calico typha cpva and calico node cpva.
```


ScheererJ (approved):
/lgtm

PR #152, opened 2022-02-21
Allow typha deployment updates even on single node clusters.

ScheererJ:

/area networking
/kind bug

Allow typha deployment updates even on single node clusters.
As calico-typha is crucial to calico-node's operation, we want it to be available all the time.
However, setting maxUnavailable=0 in the rollingUpdate strategy of the deployment will block
updates on single node clusters. This is due to the fact that calico-typha runs in the host
network and uses a static port. Hence, a new container cannot run on the single node until the
old instance is stopped.
Therefore, we bite the bullet and allow unavailability of one instance.
Fixes #

**Release note**:

```other operator
Allow calico-typha update in single node clusters
```


DockToFuture (approved):
/lgtm

PR #155, opened 2022-03-07
Removed cpu limits for calico-node, calico-typha, and the calico autoscalers and increased the memory limits

ScheererJ:

/area networking
/kind technical-debt

Removed cpu limits for calico-node, calico-typha, and the calico autoscalers (calico-node-vertical-autoscaler, calico-typha-horizontal-autoscaler, calico-typha-vertical-autoscaler) and increased the memory limits.

The memory limits of calico-node and calico-typha will no longer be autoscaler, but they are static at a fairly
high level. CPU limits are no longer enforced to prevent unnecessary throttling.
The downside of the current approach is that it will only be applied to new clusters. Existing clusters will stay
with the existing limits, which can be removed, though.

None.

**Release note**:

```other operator
Calico components will no longer have cpu limits in new clusters, calico memory limits will no longer be autoscaled, but they set to a high static value.
```

/invite @dguendisch

### DIFF
--- a/charts/internal/calico/templates/node-cpva/configmap-node-vertical-autoscaling.yaml
+++ b/charts/internal/calico/templates/node-cpva/configmap-node-vertical-autoscaling.yaml
@@ -17,17 +17,9 @@ data:
           }
         },
         "limits": {
-          "cpu": {
-            "base": "320m",
-            "step": "80m",
-            "nodesPerStep": 10,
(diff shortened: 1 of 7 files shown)


dguendisch (approved):
/lgtm
thank you!

DockToFuture (approved):
/lgtm

PR #161, opened 2022-03-15
Adapt calico-kube-controllers to remove cpu limit and increase readiness probe timeout.

ScheererJ:

/area networking
/kind bug

Adapt calico-kube-controllers to remove cpu limit and increase readiness probe timeout.

In bigger clusters with a lot of activity the calico-kube-controllers pod may run into
throttling issues due to the cpu limit. Furthermore, the readiness probe may timeout.
Fixes #

**Release note**:

```other operator
Adapt calico-kube-controllers to remove cpu limit and increase readiness probe timeout.
```


DockToFuture (approved):
/lgtm

PR #179, opened 2022-05-11
Fixed check if kube-proxy can be disabled.

ScheererJ:

/area networking
/kind bug

Fixed check if kube-proxy can be disabled.

Kube-proxy can be disabled in conjunction with calico if (and only if)
the ebpf dataplane is used. Unfortunately, the previous check did not
handle the default case, i.e. if no dataplane is configured, well. Now,
the correct default, i.e. iptables dataplane, is used.

None.

**Release note**:

```other operator

```


DockToFuture (approved):
/lgtm

PR #184, opened 2022-05-19
Increase memory limit of calico kube-controllers as new release requires more memory on large clusters

ScheererJ:

/area networking
/kind bug

Increase memory limit of calico kube-controllers as new release requires more memory on large clusters.

None.

**Release note**:

```other operator
Increased memory limit of calico kube-controllers to 2Gi.
```


DockToFuture (approved):
/lgtm

PR #204, opened 2022-08-09
Default seccomp profile for calico components

dimityrmirchev:

/area security
/kind enhancement

This PR adds a seccomp profile to calico workload running in the shoot.
It also allows `/proc` as allowed host path in the psp definition for `calico-node`.

Fixes #

**Release note**:

```other operator
Pods for `calico-kube-controllers`, `calico-node`, `calico-node-vertical-autoscaler`, `calico-typha`, `calico-typha-horizontal-autoscaler` and `calico-typha-vertical-autoscaler` components now have their seccomp profile set to "RuntimeDefault". 
```

### DIFF
--- a/charts/internal/calico/templates/node/daemonset-calico-node.yaml
+++ b/charts/internal/calico/templates/node/daemonset-calico-node.yaml
@@ -45,6 +45,11 @@ spec:
         - effect: NoExecute
           operator: Exists
       serviceAccountName: calico-node
+      {{- if semverCompare ">= 1.19" .Capabilities.KubeVersion.GitVersion }}
+      securityContext:
+        seccompProfile:
+          type: RuntimeDefault
+      {{- end }}
       # Minimize downtime during a rolling upgrade or deletion; tell Kubernetes to do a "force
       # deletion": https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods.
       terminationGracePeriodSeconds: 0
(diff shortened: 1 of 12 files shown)


charts/internal/calico/templates/node/daemonset-calico-node.yaml
```diff
@@ -45,6 +45,11 @@ spec:
         - effect: NoExecute
           operator: Exists
       serviceAccountName: calico-node
+      {{- if semverCompare ">= 1.19" .Capabilities.KubeVersion.GitVersion }}
+      securityContext:
+        seccompProfile:
+          type: RuntimeDefault
+      {{- end }}
```
ScheererJ, line 52: Why do you specify a `seccompProfile` for `calico-node` when all of its containers have `privileged: true`?
According to the [documentation](https://kubernetes.io/docs/tutorials/security/seccomp/#before-you-begin) of `seccompProfile`, this has no effect.

ScheererJ (changes requested):

charts/internal/calico/templates/node/daemonset-calico-node.yaml
dimityrmirchev: Yes, this is right. Basically this does nothing with the current setup of `calico-node` because the single container runs as privileged. However it acts as guardrails against:
1. adding a new container to the pod without seccomp profile
2. removing the `privileged: true` from the container and still running as `Unconfined`

ScheererJ: Do you plan on proposing those changes in the upstream calico repository as well? It feels shaky if we add it here without having upstream commitment.

dimityrmirchev: You are right. I will check if they have any ongoing (or planned) works in this direction and if not I will propose the changes I made here.

dimityrmirchev:
As discussed I also proposed the changes in the upstream projectcalico/calico#6524

ScheererJ (approved):
/lgtm

PR #209, opened 2022-08-26
Add an option to run calico-node in non-privileged mode

ialidzhikov:
/area networking
/kind enhancement

The PR adds an option to run calico-node in non-privileged mode. For more details, see `docs/usage-as-operator.md`

The implementation is heavily inspired from the [Tigera Calico operator](https://github.com/tigera/operator).
Along the way I had to exclude and do not render the ebpf dataplane init container and ebpf related volumes/mounts when ebfp dataplane is not enabled (as done in the Tigera Calico operator). Previously they were rendered unconditionally - no matter ebpf is enabled or not.
The motivation for this is that in non-privileged mode calico-node does not support ebpf dataplane.

**Release note**:

```feature operator
A new feature gate named `NonPrivilegedCalicoNode` is now added to the networking-calico extension. It allows running the long-lived calico-node container in non-privileged and non-root mode.
```


ScheererJ (approved):
/lgtm

PR #222, opened 2022-11-11
Adapt high-availability configuration

timuthy:

/area networking
/area high-availability
/kind enhancement

This PR prepares the following components for HA by labeling them with the corresponding `high-availability-config.resources.gardener.cloud/type` and defining the required `PodDisruptionBudget`, please see [this document](https://github.com/gardener/gardener/blob/master/docs/development/high-availability.md) for more information.
- Extension Admission Controller
- Extension Controller

In addition, it increases the nodes-to-replica ration for `Typha`, so that the cluster will have 2 replicas as soon as there are two nodes.

Part of https://github.com/gardener/gardener/issues/6529

Remaining components related to `CPA` and `CPVA` don't offer any leader election, so for the time being they'll remain running with a single replica only. 

**Release note**:

```feature operator
Several seed components (namely `Extension Controller`, `Extension Admission Controller`) have been adjusted to the high availability configuration supported by Gardener as of release `v1.60`. See [this document](https://github.com/gardener/gardener/blob/master/docs/development/high-availability.md) for more information about HA in gardener.
```
```feature user
The high availability configuration for `Typha` has been revised: For failure tolerance, at least two `Typha` instances are deployed as soon as the cluster has two or more nodes.
```


charts/internal/calico/templates/typha-cpha/configmap-typha-horizontal-autoscaling.yaml
```diff
@@ -12,7 +12,7 @@ data:
       "nodesToReplicas":
       [
         [1, 1],
-        [10, 2],
+        [2, 2],
```
ScheererJ, line 15: Did you test this on a two node cluster? I hope that `maxUnavailable = 1` should allow a rolling update, but this requires on a two node cluster that an existing typha instance is shut down before a new one is created. Otherwise, the new one will stay in pending forever.

ScheererJ (changes requested):
Please also check the `verify` error.

rfranzke (commented):
lgtm from my side

charts/internal/calico/templates/typha-cpha/configmap-typha-horizontal-autoscaling.yaml
timuthy: Hm, isn't the `calico-typha-horizontal-autoscaler` Deployment rolled because the ConfigMap has changed? Because this deployment doesn't have a PDB.
Nevertheless, I tried out changing the config as shown above and a second `calico-typha-deploy` Pod was created, the first one kept running.

charts/gardener-extension-networking-calico/templates/deployment.yaml
```diff
@@ -1,19 +1,17 @@
...
 metadata:
-  name: gardener-extension-networking-calico
+  name: {{ include "name" . }}
   namespace: {{ .Release.Namespace }}
   labels:
-    app.kubernetes.io/name: gardener-extension-networking-calico
-    helm.sh/chart: gardener-extension-networking-calico
-    app.kubernetes.io/instance: {{ .Release.Name }}
+{{ include "commonObjectLabels" . | indent 4 }}
+    high-availability-config.resources.gardener.cloud/type: controller
```
ScheererJ, line 8: Why is this extension of type controller while the cilium extension is modelled as server?

charts/internal/calico/templates/typha-cpha/configmap-typha-horizontal-autoscaling.yaml
ScheererJ: My question was targeting the `calico-typha` deployment. The question was rather did the rollout of an update works on a two node cluster.

ScheererJ (changes requested):

charts/gardener-extension-networking-calico/templates/deployment.yaml
rfranzke: This extension does not serve any webhooks while Cilium does.

ScheererJ (approved):
/lgtm

PR #224, opened 2022-12-01
omit backend field when empty

DockToFuture:

/area networking
/kind bug

Backend field is omitted when not specified.
This will avoid setting `backend: null` when the shoot mutator webhook in the infrastructure extension is called which might lead to error during validation.

```
error: error validating "STDIN": error validating data: unknown object type "nil" in Shoot.spec.networking.providerConfig.backend; if you choose to ignore these errors, turn validation off with --validate=false
```

Fixes #

**Release note**:

```other operator
Backend field is omitted when not specified.
```


ScheererJ (approved):
/lgtm

PR #228, opened 2022-12-23
fix deployment

DockToFuture:

/area networking
/kind bug

This PR fixes the broken symlink for the landscape deployment. The folder where the symlink points to is not in the providerConfig and can't thus be resolved.

Fixes #

**Release note**:

```other operator
none
```


ScheererJ (approved):
/lgtm

PR #229, opened 2023-01-12
Disable SNAT for clusters without overlay

DockToFuture:

/area networking
/kind enhancement

This PR adds an additional calico ippool if running without overlay to disable SNAT within the node CIDR range.
When switching from `overlay` mode to `non-overlay` mode the ippool is added before bird is disabled, which might lead to unreachable or incompatible routes. Thus an init container was added to calico node to clean up those incompatible routes. Furthermore are also old routes cleaned up which were created beforehand by bird and are of no use anymore.
Fixes #

Tested on:

- [x] aws
- [x] gcp
- [x] alicloud
- [x] openstack
- [x] azure

**Release note**:

```other operator
SNAT is disabled for clusters without overlay. Incompatible routes created by bird process when switching the overlay mode are now properly cleaned up by an init container.
```


ScheererJ (dismissed):
/lgtm

ScheererJ (approved):
/lgtm

PR #231, opened 2023-01-17
Drop nodeSelector and tolerations from Shoot System Components

timuthy:

/area networking
/kind cleanup

After the integration of https://github.com/gardener/gardener/pull/7204, this PR cleans up relevant deployments of Shoot System Components:
- removes `nodeSelector`
- removes all `CriticalAddonsOnly` tolerations (not related to [#7204](https://github.com/gardener/gardener/pull/7204) but removed along the way)
- adds `NoExecute` toleration for Calico-Node as this pod is supposed to run at all times or it doesn't make much sense to evict it (please double check @ScheererJ @DockToFuture)

Please see the individual commit messages for more information.

Part of https://github.com/gardener/gardener/issues/7103

~Remains in draft until Gardener `v1.63` is released.~

**Release note**:

```breaking operator
Node selectors and general tolerations were removed from Calico deployments applied to the shoot cluster in favor of Gardener's automatic handling added in https://github.com/gardener/gardener/pull/7204.
This version of the Calico extension expects a minimum Gardener version `v1.63`.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,13 @@ Finally after successful reconciliation an output similar to the one below shoul
       kind: NetworkStatus
 ```
 
+## Compatibility
+
+The following lists known compatibility issues of this extension controller with other Gardener components.
+
+| Calico Extension | Gardener | Action | Notes |
+| ------------- | -------- | ------ |  --- |
+| `>= v1.28.0` | `< v1.63.0` | Please first update Gardener components to `>= v1.63.0`. | Without the mentioned minimum Gardener version, Calico `Pod`s are not only scheduled to dedicated system component nodes in the shoot cluster.
 ----
 
 ## How to start using or developing this extension controller locally
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ The following table lists known compatibility issues of this extension controlle
 
 | Calico Extension | Gardener | Action | Notes |
 | ------------- | -------- | ------ |  --- |
-| `>= v1.28.0` | `< v1.63.0` | Please first update Gardener components to `>= v1.63.0`. | Without the mentioned minimum Gardener version, Calico `Pod`s are not only scheduled to dedicated system component nodes in the shoot cluster.
+| `>= v1.30.0` | `< v1.63.0` | Please first update Gardener components to `>= v1.63.0`. | Without the mentioned minimum Gardener version, Calico `Pod`s are not only scheduled to dedicated system component nodes in the shoot cluster.
 ----
 
 ## How to start using or developing this extension controller locally
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ Finally after successful reconciliation an output similar to the one below shoul
 
 ## Compatibility
 
-The following lists known compatibility issues of this extension controller with other Gardener components.
+The following table lists known compatibility issues of this extension controller with other Gardener components.
 
 | Calico Extension | Gardener | Action | Notes |
 | ------------- | -------- | ------ |  --- |
--- a/charts/internal/calico/templates/typha/deployment-calico-typha.yaml
+++ b/charts/internal/calico/templates/typha/deployment-calico-typha.yaml
@@ -53,11 +53,6 @@ spec:
       # Make sure typha gets scheduled on all nodes.
       - effect: NoSchedule
         operator: Exists
-      # Mark the pod as a critical add-on for rescheduling.
-      - key: CriticalAddonsOnly
-        operator: Exists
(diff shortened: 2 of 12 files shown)


rfranzke (commented):
Can you add something like https://github.com/gardener/gardener-extension-provider-aws#compatibility here as well?

timuthy:
Good idea @rfranzke, will do.

charts/internal/calico/templates/typha/deployment-calico-typha.yaml
```diff
@@ -45,19 +45,10 @@ spec:
...
-      {{- end }}
       tolerations:
       # Make sure typha gets scheduled on all nodes.
       - effect: NoSchedule
         operator: Exists
-      # Mark the pod as a critical add-on for rescheduling.
-      - key: CriticalAddonsOnly
-        operator: Exists
-      - effect: NoExecute
-        operator: Exists
```
ScheererJ, line 60: Not sure about this one. `calico-typha` is even more critical than `calico-node` in that the latter will not work without the former. However, we should have two replicas starting with two nodes since https://github.com/gardener/gardener-extension-networking-calico/pull/222. Therefore, it should be fine to reschedule a `calico-typha` replica.
Nevertheless, it should run almost all of the time. Is the `NoExecute` toleration required for this?

ScheererJ (commented):
Looks good overall.
Compatibility table as suggested by @rfranzke would certainly be a great addition.

charts/internal/calico/templates/typha/deployment-calico-typha.yaml
timuthy: I think it is required but please keep in mind that the Kube-Apiserver automatically adds tolerations for `node.kubernetes.io/not-ready` and `node.kubernetes.io/unreachable` with `tolerationSeconds=300` (see [taint based evictions](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#taint-based-evictions))
The benefit of removing a ever lasting toleration like the one above is, that taint based evictions finally work for the Typha replicas as well, e.g. if a node is not ready for `> 300s` then the Typha pod is evicted. WDYT?

ScheererJ: For the case that some random node is temporarily not ready, I agree that it makes a lot of sense that `calico-typha` gets evicted and rescheduled assuming that a second replica is happily running on a working node.
However, if all nodes are not ready for over 5 minutes I am not sure what will happen. Does a cluster then have a chance of ever recovering or will at least some nodes be replaced? The reasoning is as follows:
1. After 5 minutes `calico-typha` will be evicted of all nodes and there will be no nodes available for rescheduling.
2. `calico-node` will be rather unhappy if no `calico-typha` instance is running.
3. Not exactly sure what will happen then.

WDYT?

timuthy: That's a good point and conceptually it should recover. However, let me reproduce this scenario and get back with the results.

timuthy: I can confirm that the procedure works as expected and the cluster recovered completely after no `Node`s were ready, `calico-typha` was evicted. After the `Node`s were `Ready` again, `calico-typha` was scheduled and everything returned to a normal state again.

README.md
```diff
@@ -56,6 +56,13 @@ Finally after successful reconciliation an output similar to the one below shoul
       kind: NetworkStatus
 ```
 
+## Compatibility
+
+The following table lists known compatibility issues of this extension controller with other Gardener components.
+
+| Calico Extension | Gardener | Action | Notes |
+| ------------- | -------- | ------ |  --- |
+| `>= v1.28.0` | `< v1.63.0` | Please first update Gardener components to `>= v1.63.0`. | Without the mentioned minimum Gardener version, Calico `Pod`s are not only scheduled to dedicated system component nodes in the shoot cluster.
```
ScheererJ, line 65: I suppose this needs to be updated.

```suggestion
| `>= v1.30.0` | `< v1.63.0` | Please first update Gardener components to `>= v1.63.0`. | Without the mentioned minimum Gardener version, Calico `Pod`s are not only scheduled to dedicated system component nodes in the shoot cluster.
```

ScheererJ (changes requested):
Could you please recheck the version constraint? I think we created some releases meanwhile.

ScheererJ (approved):
/lgtm

PR #244, opened 2023-02-22
update docs

DockToFuture:

/area networking
/kind enhancement

Documentation is updated to correctly describe now how the overlay is enabled and disabled.

Fixes #

**Release note**:

```other operator
none
```

### DIFF
--- a/example/20-network.yaml
+++ b/example/20-network.yaml
@@ -46,11 +46,14 @@ spec:
   providerConfig:
     apiVersion: calico.networking.extensions.gardener.cloud/v1alpha1
     kind: NetworkConfig
-    backend: bird
-    ipam:
-      type: host-local
-      cidr: usePodCIDR
-#   ipAutodetectionMethod: interface=eth0
-#   ipip: CrossSubnet
+    overlay:
+      enabled: false
+    snatToUpstreamDNS:
+      enabled: true
+#   ebpfDataplane:
+#     enabled: true
 #   typha:
 #     enabled: true
+#   ipam:
+#     type: host-local
+#     cidr: usePodCIDR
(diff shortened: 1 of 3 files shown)


example/20-network.yaml
```diff
@@ -46,11 +46,15 @@ spec:
...
-    ipam:
-      type: host-local
-      cidr: usePodCIDR
+    overlay:
+      enabled: false
+    snatToUpstreamDNS:
+      enabled: true
+#   ebpfDataplane:
+#     enabled: true
 #   ipAutodetectionMethod: interface=eth0
```
ScheererJ, line 55: Can you please also remove this entry? It seems deprecated and should not be included in the example anymore.

example/20-network.yaml
```diff
@@ -46,11 +46,15 @@ spec:
...
-#   ipip: CrossSubnet
 #   typha:
 #     enabled: true
+#   ipam:
+#     type: host-local
+#     cidr: usePodCIDR
```
ScheererJ, line 60: Nit: Any particular reason why you moved `ipam` to the bottom?

ScheererJ (changes requested):
Thanks. Looks good mostly.

example/20-network.yaml
DockToFuture: I've never seen that anybody has set this field and do we really want stakeholder to set this?

ScheererJ: Ah, I did not get that you commented it. For sure, this is expert configuration, which should not necessarily be set. No issue to keep it commented out.

ScheererJ (approved):
/lgtm

PR #249, opened 2023-03-15
Allow non-specified node cidr.

ScheererJ:

/area networking
/kind bug

Allow non-specified node cidr.

The changes related to source network address translation unintentionally enforced a specified node cidr. This change relaxes the condition, i.e. node cidr is optional again.
The tests have been refactored to make it easier to add new variants without node cidr specified.

None.

**Release note**:

```other operator
Calico extension does not crash anymore when node cidr is not specified.
```


DockToFuture (approved):
/lgtm

PR #266, opened 2023-05-02
Induce component version using ldflags

shafeeqes:

/area networking
/kind enhancement

Currently, it is hard to find the version of the `networking-calico` images.
This PR improves this by inducing the version using ldflags.
Version can be printed by using the `--version` flag.
eg:
```
$ docker run eu.gcr.io/gardener-project/gardener/extensions/networking-calico:v1.33.0-dev-390632cfcd97607832fbae0c0db858e1d9adccb7-linux-arm64 --version
gardener-extension v1.33.0-dev-390632cfcd97607832fbae0c0db858e1d9adccb7
```
Also we log the version during the starting of the component.

Also with this PR, docker images built by make docker-images are tagged and build with a version including the commit hash (exactly like the images built by the pipeline).

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,10 @@ FROM golang:1.20.2 AS builder
 
 WORKDIR /go/src/github.com/gardener/gardener-extension-networking-calico
 COPY . .
-RUN make install
+
+ARG EFFECTIVE_VERSION
+
+RUN make install EFFECTIVE_VERSION=$EFFECTIVE_VERSION
 
 ############# gardener-extension-networking-calico
 FROM gcr.io/distroless/static-debian11:nonroot AS gardener-extension-networking-calico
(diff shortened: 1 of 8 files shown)


ScheererJ (commented):
Please do not get me wrong. I am not opposed to the change.
However, if the goal is to "just" get the version why not just open the container image link in a browser. It usually includes the version tag.

shafeeqes:
> Please do not get me wrong. I am not opposed to the change. However, if the goal is to "just" get the version why not just open the container image link in a browser. It usually includes the version tag.

Hmm.. That also works. But there isn't any harm in one more way of finding the version, right?

ScheererJ (approved):
/lgtm

PR #271, opened 2023-05-16
topology-aware routing: Adapt to the upstream annotation change in K8s 1.27

ialidzhikov:

/area high-availability
/area networking
/kind enhancement

See https://github.com/gardener/gardener/pull/7933

N/A

N/A

**Release note**:

```other operator
NONE
```

### DIFF
--- a/charts/gardener-extension-admission-calico/charts/runtime/templates/service.yaml
+++ b/charts/gardener-extension-admission-calico/charts/runtime/templates/service.yaml
@@ -5,9 +5,13 @@ metadata:
   namespace: {{ .Release.Namespace }}
   annotations:
     networking.resources.gardener.cloud/from-all-webhook-targets-allowed-ports: '[{"protocol":"TCP","port":{{ .Values.global.webhookConfig.serverPort }}}]'
-  {{- if .Values.global.service.topologyAwareRouting.enabled }}
+    {{- if .Values.global.service.topologyAwareRouting.enabled }}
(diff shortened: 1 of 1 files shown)


ScheererJ (approved):
/lgtm

ialidzhikov:
> /lgtm, btw is there also a reason why the annotation is not set here: https://github.com/gardener/gardener-extension-networking-calico/blob/master/charts/gardener-extension-networking-calico/templates/service.yaml

Yes, there is. Long story short, we make Service topology-aware when there is a need for it. For example the networking-cilium extension's Service is topology-aware as it serves a Shoot webhook that mutates the node-local-dns DaemonSet and ConfigMap. Hence, in a Seed cluster the networking-cilium Service is consumed by kube-apiserver Pods (for the webhook communication).
networking-calico does not run such Shoot webhook, hence it is not consumed by other Seed components.
See https://github.com/gardener/gardener/issues/6718 - in the issue description I tried to maintain similar reasons on some why Services can (have to) topology-aware. There is also https://github.com/gardener/gardener/blob/master/docs/usage/topology_aware_routing.md which documents other components + reasoning, and contains docs about prerequisites for a Service to be topology-aware.
Let me know if I missed something or you have something else in mind.

DockToFuture:
> ConfigMap. Hence, in a Seed cluster the networking-cilium Service is consumed by kube-apiserver Pods (for the webhook communication). The networking-cilium Service can benefit from the topology-aware routing feature as it is consumed by clients (the kube-apiserver Pods) from the same cluster.

Got it, thanks!

PR #285, opened 2023-07-28
Allow propagating pod routes to nodes without overlay network

ScheererJ:

/area networking
/kind enhancement
/kind cleanup

Allow propagating pod routes to nodes without overlay network.

Some infrastructures may implement routing in the virtual network by moving all traffic via a routing component. This can introduce additional latency and will eventually become a bottleneck. In case the network is sufficiently capable of using a "short-cut" by applying local routes on the nodes, it can be desirable to do so.

This change allows to trigger propagation of pod routes to all nodes via an option in the networking configuration.

None.

**Release note**:

```other operator
Allow propagating pod routes to nodes without overlay network by specifying `shoot.spec.networking.providerConfig.overlay.createPodRoutes: true`
```


pkg/controller/actuator_reconcile.go
```diff
@@ -110,6 +110,13 @@ func (a *actuator) Reconcile(ctx context.Context, _ logr.Logger, network *extens
 			networkConfig.IPv4.Mode = (*calicov1alpha1.IPv4PoolMode)(pointer.String(string(calicov1alpha1.Never)))
 			networkConfig.Backend = (*calicov1alpha1.Backend)(pointer.String(string(calicov1alpha1.None)))
 		}
+		if networkConfig.Overlay != nil && networkConfig.Overlay.CreatePodRoutes != nil {
+			if *networkConfig.Overlay.CreatePodRoutes {
+				networkConfig.Backend = (*calicov1alpha1.Backend)(pointer.String(string(calicov1alpha1.Bird)))
+			} else {
+				networkConfig.Backend = (*calicov1alpha1.Backend)(pointer.String(string(calicov1alpha1.None)))
+			}
+		}
```
axel7born, line 119: Here it is possible to disable `bird` even if `Overlay` is enabled. 
Is that a valid configuration?
The logic in the related Cilium change seems to be different.

ScheererJ: Yes, indeed. What do you propose?

DockToFuture: Maybe something like that:

```
if networkConfig.Overlay != nil && !networkConfig.Overlay.Enabled && networkConfig.Overlay.CreatePodRoutes != nil && networkConfig.Overlay.CreatePodRoutes{
networkConfig.Backend = (*calicov1alpha1.Backend)(pointer.String(string(calicov1alpha1.Bird)))
}
```

pkg/controller/actuator_reconcile.go
```diff
@@ -110,6 +110,13 @@ func (a *actuator) Reconcile(ctx context.Context, _ logr.Logger, network *extens
 			networkConfig.IPv4.Mode = (*calicov1alpha1.IPv4PoolMode)(pointer.String(string(calicov1alpha1.Never)))
 			networkConfig.Backend = (*calicov1alpha1.Backend)(pointer.String(string(calicov1alpha1.None)))
```
DockToFuture, line 111: Maybe like that:
```suggestion
if networkConfig.Overlay.CreatePodRoutes != nil && networkConfig.Overlay.CreatePodRoutes{ 
   networkConfig.Backend = (*calicov1alpha1.Backend)(pointer.String(string(calicov1alpha1.Bird))) 
} else {
			networkConfig.Backend = (*calicov1alpha1.Backend)(pointer.String(string(calicov1alpha1.None)))
			}
```

DockToFuture:
Maybe you can also mention the field how this can be configured in the release notes.

pkg/controller/actuator_reconcile.go
ScheererJ: @DockToFuture @axel7born Is this how you expected it?

DockToFuture (approved):
/lgtm

PR #293, opened 2023-08-23
Vendor `gardener@1.77.2`

shafeeqes:

/area open-source
/kind enhancement

Vendor `gardener@1.77.2` and adapt breaking changes.

Fixes #

/cc @dimityrmirchev for https://github.com/gardener/gardener-extension-networking-calico/pull/293/commits/32a9e199190a723a7ffbe3f9c40459442bc2ff4c

**Release note**:

```other operator
The following dependency is updated:
- github.com/gardener/gardener: v1.76.0 -> v1.77.1
```


dimityrmirchev:
lgtm for https://github.com/gardener/gardener-extension-networking-calico/commit/32a9e199190a723a7ffbe3f9c40459442bc2ff4c

I also managed to notice a typo in the release note. It should say that g/g was updated to 1.77.1 instead of v1.71.1

dimityrmirchev:
I think that we can also apply the following patch


```patch
diff --git a/pkg/controller/actuator_reconcile.go b/pkg/controller/actuator_reconcile.go
index 5ea64028..c08bf0de 100644
--- a/pkg/controller/actuator_reconcile.go
+++ b/pkg/controller/actuator_reconcile.go
@@ -17,13 +17,14 @@ package controller
 import (
 	"context"
 	"fmt"
+	"time"
 
 	extensionscontroller "github.com/gardener/gardener/extensions/pkg/controller"
-	"github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
 	gardencorev1beta1helper "github.com/gardener/gardener/pkg/apis/core/v1beta1/helper"
 	extensionsv1alpha1 "github.com/gardener/gardener/pkg/apis/extensions/v1alpha1"
 	gardenerkubernetes "github.com/gardener/gardener/pkg/client/kubernetes"
 	"github.com/gardener/gardener/pkg/utils/chart"
+	"github.com/gardener/gardener/pkg/utils/managedresources"
 	"github.com/gardener/gardener/pkg/utils/managedresources/builder"
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
@@ -148,17 +149,14 @@ func (a *actuator) Reconcile(ctx context.Context, _ logr.Logger, network *extens
 		return err
 	}
 
-	secret, secretRefs := calicoSecret(a.client, calicoChart, network.Namespace)
-	if err := secret.Reconcile(ctx); err != nil {
+	data := map[string][]byte{charts.CalicoConfigKey: calicoChart}
+	if err := managedresources.CreateForShoot(ctx, a.client, network.Namespace, CalicoConfigSecretName, "extension-networking-calico", false, data); err != nil {
 		return err
 	}
 
-	if err := builder.
-		NewManagedResource(a.client).
-		WithNamespacedName(network.Namespace, CalicoConfigSecretName).
-		WithSecretRefs(secretRefs).
-		WithInjectedLabels(map[string]string{constants.ShootNoCleanup: "true"}).
-		Reconcile(ctx); err != nil {
+	timeoutSeedCtx, cancelSeedCtx := context.WithTimeout(ctx, 2*time.Minute)
+	defer cancelSeedCtx()
+	if err := managedresources.WaitUntilHealthy(timeoutSeedCtx, a.client, network.Namespace, CalicoConfigSecretName); err != nil {
 		return err
 	}
 

```

shafeeqes:
@dimityrmirchev The wait for MR cannot be added, because the MR contains a deployment, and the nodes are only deployed after the network extension, so they cannot get healthy at this point of time. The e2e tests were failing, so I pushed a commit reverting this change.

pkg/controller/actuator_reconcile.go
```diff
@@ -139,12 +138,6 @@ func (a *actuator) Reconcile(ctx context.Context, _ logr.Logger, network *extens
 		return err
 	}
 
-	timeoutCtx, cancelCtx := context.WithTimeout(ctx, 2*time.Minute)
-	defer cancelCtx()
-	if err := managedresources.WaitUntilHealthy(timeoutCtx, a.client, network.Namespace, CalicoConfigManagedResourceName); err != nil {
-		return err
-	}
```
ScheererJ, line 163: Could you please clarify why you removed this?

ScheererJ (commented):
Thanks for the change. It looks good overall. I just have one question with regards code you removed.

pkg/controller/actuator_reconcile.go
shafeeqes: I added it as per https://github.com/gardener/gardener-extension-networking-calico/pull/293#issuecomment-1689347271, but as explained in https://github.com/gardener/gardener-extension-networking-calico/pull/293#issuecomment-1700483791, the MR cannot be healthy until the deployment is ready.

ScheererJ: Ok, now I get it. This was only introduced with this PR and then removed again.
Thanks for pointing it out.

Works for me.

ScheererJ (approved):
/lgtm

PR #297, opened 2023-09-22
Fix selector of cluster-proportional autoscalers for calico-typha.

ScheererJ:

/area networking
/kind bug

Fix selector of cluster-proportional autoscalers for calico-typha.

Both, the vertical as well as the horizontal autoscalers used the same label. This causes issues when other controllers try to validate the the correct amount of pods are running as both vertical as well as horizontal autoscalers match the label selector.

None.

**Release note**:

```other operator
Vertical and horizontal cluster-proportional autoscalers for calico-typha now use different label selectors.
```


rfranzke (approved):
/lgtm

PR #301, opened 2023-09-28
[GEP-21] provider-local single-stack ipv6 implementation

nschad:
/area networking
/kind enhancement
/kind api-change

See https://github.com/gardener/gardener/pull/8574

Part of https://github.com/gardener/gardener/issues/7051

**Release note**:

```feature developer
add ipv6 support for this extension
```


pkg/apis/calico/types_network.go
```diff
@@ -26,38 +26,51 @@ const (
...
+	PoolIPIP  Pool = "ipip"
+	PoolVXLan Pool = "vxlan"
 )
 
 // IPv4 contains configuration for calico ipv4 specific settings
 type IPv4 struct {
 	// Pool configures the type of ip pool for the tunnel interface.
 	// https://docs.projectcalico.org/v3.8/reference/node/configuration#environment-variables
-	Pool *IPv4Pool
+	Pool *Pool
```
axel7born, line 51: Do we really change the API here?

pkg/apis/calico/v1alpha1/types_network.go
```diff
@@ -26,41 +26,59 @@ const (
...
 
+// IPv6 contains configuration for calico ipv6 specific settings
+type IPv6 struct {
+	// Pool configures the type of ip pool for the tunnel interface
+	// https://docs.tigera.io/calico/latest/reference/configure-calico-node#configuring-the-default-ip-pools
+	// +optional
+	Pool *Pool `json:"pool,omitempty"`
+	// Mode is the mode for the IPv6 Pool (e.g. Always, Never, CrossSubnet)
+	// TODO: VXLAN also supports CrossSubnet for VXLAN. Why is this not supported?
+	// vxlan pools accept only Always and Never (unchecked)
```
axel7born, line 73: Though, I don't know the answer to the TODO question, I would prefer not to have it here in the api-reference.
You could open an issue regarding the CrossSubnet support?

pkg/apis/calico/types_network.go
nschad: No it's basically just renamed. The only new thing for the API is the IPv6 stuff

pkg/apis/calico/v1alpha1/types_network.go
nschad: Absolutely the TODO makes no sense here. I will create an issue about it

pkg/apis/calico/types_network.go
nschad: Ah okay I see. I figured in a IPV4 struct having the PoolMode to be named `IPv4PoolMode` is just word clutter. It also deduped some code and uses less "bloat".

However I don't mind reverting

timebertt: I would vote for keeping the short type names (without `IPv4`) as it fits the use across IPv4 and IPv6.
To keep the go API package backward-compatible, I suggest adding type aliases for `IPv4{Pool,PoolMode}`. That should be good enough :)

timebertt:
PR is rebased and all comments addressed. PTAL @axel7born :)

pkg/apis/calico/types_network.go
axel7born: OK, I'm fine with this.

pkg/controller/actuator_reconcile.go
```diff
@@ -82,19 +84,29 @@ func (a *actuator) Reconcile(ctx context.Context, _ logr.Logger, network *extens
...
+			}
+			networkConfig.IPv6.AutoDetectionMethod = &autodetectionMode
 		}
-		networkConfig.IPv4.AutoDetectionMethod = &autodetectionMode
 	}
 
 	if networkConfig != nil {
 		if networkConfig.Overlay != nil && networkConfig.Overlay.Enabled {
-			networkConfig.IPv4.Mode = (*calicov1alpha1.IPv4PoolMode)(pointer.String(string(calicov1alpha1.Always)))
+			networkConfig.IPv4.Mode = (*calicov1alpha1.PoolMode)(pointer.String(string(calicov1alpha1.Always)))
```
DockToFuture, line 105:
```suggestion
			if ipFamilies.Has(extensionsv1alpha1.IPFamilyIPv4) {
				networkConfig.IPv4.Mode = (*calicov1alpha1.PoolMode)(pointer.String(string(calicov1alpha1.Always)))
			}
			if ipFamilies.Has(extensionsv1alpha1.IPFamilyIPv6) {
				networkConfig.IPv6.Mode = (*calicov1alpha1.PoolMode)(pointer.String(string(calicov1alpha1.Always)))
			}
```

pkg/controller/actuator_reconcile.go
```diff
@@ -82,19 +84,29 @@ func (a *actuator) Reconcile(ctx context.Context, _ logr.Logger, network *extens
...
 			networkConfig.Backend = (*calicov1alpha1.Backend)(pointer.String(string(calicov1alpha1.Bird)))
 		}
 		if networkConfig.Overlay != nil && !networkConfig.Overlay.Enabled {
-			networkConfig.IPv4.Mode = (*calicov1alpha1.IPv4PoolMode)(pointer.String(string(calicov1alpha1.Never)))
+			networkConfig.IPv4.Mode = (*calicov1alpha1.PoolMode)(pointer.String(string(calicov1alpha1.Never)))
```
DockToFuture, line 109:
```suggestion
			if ipFamilies.Has(extensionsv1alpha1.IPFamilyIPv4) {
				networkConfig.IPv4.Mode = (*calicov1alpha1.PoolMode)(pointer.String(string(calicov1alpha1.Never)))
			}
			if ipFamilies.Has(extensionsv1alpha1.IPFamilyIPv6) {
				networkConfig.IPv6.Mode = (*calicov1alpha1.PoolMode)(pointer.String(string(calicov1alpha1.Never)))
			}
```

nschad:
Hey @DockToFuture,

We have implemented the suggestions can you please PTAL?

DockToFuture:
> Hey @DockToFuture,
> 
> We have implemented the suggestions can you please PTAL?

The PR looks good to me, we will test it once more, to validate that it does not break anything and then get it in. :)

pkg/controller/actuator_reconcile.go
DockToFuture: Any reasons why this got now removed?

DockToFuture:
> Hey @DockToFuture,
> 
> We have implemented the suggestions can you please PTAL?

Any reasons why this got now removed?

nschad:
> > Hey @DockToFuture,
> > We have implemented the suggestions can you please PTAL?
> 
> Any reasons why this got now removed?

I messed up.

pkg/charts/utils.go
```diff
@@ -204,10 +215,57 @@ func ComputeCalicoChartValues(
...
+			AutoDetectionMethod: nil,
+		}
+	}
+
+	if isIPv6 {
+		c.IPAM.AssignIPv6 = true
+		c.IPAM.Subnet = usePodCIDRv6
+		// NOTE: calico guide (https://docs.tigera.io/calico/latest/networking/ipam/ipv6#enable-ipv6-only) requires ipam
+		// type to be calico-ipam for single IPv6 or dual-stack clusters.
+		c.IPAM.IPAMType = calicoIPAM
```
nschad, line 245: Did we interpret this correctly? @DockToFuture

nschad: CC: @axel7born

nschad:
Hey,

FYI: We dropped one PreFactor commit because it caused the E2E test to fail. We accidentally changed the behavior in this dropped commit from sometimes deploy `IPPool` to **always** deploy `IPPool` which causes the `IPPool` to be created by the extension with `IPIPMode: Never` which then causes the pod-network ping test to fail. 

The behavior is now back to how it was and the ippool is not deployed anymore  by the extension, instead it's created by calico itself.

pkg/charts/utils.go
DockToFuture: Yes, unfortunately it's currently not possible to use the host-local IPAM plugin. I've raised an issue in regards to that: https://github.com/projectcalico/calico/issues/8190

DockToFuture (approved):
lgtm

PR #394, opened 2024-05-06
[GEP-19] Adapt monitoring configuration

rfranzke:

/area monitoring
/kind enhancement

https://github.com/gardener/gardener/pull/9695 changes the contract of how extension can provide monitoring configuration, see https://github.com/gardener/gardener/pull/9695/files#diff-93972179f36fa195b948f0f4979acab75697f99f3c51ab889994d9d8ba9f170b. This PR adapts this extension accordingly. It also cleans up no longer needed configuration (if found).

Part of https://github.com/gardener/gardener/issues/9065

/cc @ScheererJ 

**Release note**:

```other operator
This extension is now using the new way of providing monitoring configuration (ref [GEP-19](https://github.com/gardener/gardener/blob/master/docs/proposals/19-migrating-observability-stack-to-operators.md)) in case a shoot cluster's Prometheus has been migrated to management via `prometheus-operator`.
```


axel7born:
Testing the old behavior, I've seen that the calico dashboards are not added anymore.
However, the reason seems to be this [change](https://github.com/gardener/gardener/pull/9624) to g/g.
It seems that only `dashboard_users` and `dashboard_operators` are added.
https://github.com/gardener/gardener/blob/96156b1d9d91d1982ca45bafe8ede353170402e6/pkg/component/observability/plutono/plutono.go#L902

rfranzke:
@axel7born Can you explain more? What exactly do you mean by "testing the old behavior"? What exactly is not working that was working before this PR? 👀

axel7born:
@rfranzke By "old behavior" I meant `.Values.gep19Monitoring: false`. Dashboards from `ConfigMaps` with the annotation `dashboard.monitoring.gardener.cloud/shoot: "true"` should be added to the ConfigMap `plutono-dashboards`. That doesn't happen anymore.

rfranzke:
Hm, no, they shouldn't be added anymore to this `ConfigMap`. The sidecar in the plutono pod dynamically loads the content of the dashboards from the `ConfigMap` and writes it into a local file shared with the main plutono container. Hence, `gardenlet`'s `plutono-dashboards` is no longer augmented with dashboards from the extensions.

Did you check whether your dashboard is available in Plutono?

axel7born:
It's not available in Plutono. Is there something else required to get the dashboard?

rfranzke:
Thanks for notifying, I think that's a bug that was introduced by https://github.com/gardener/gardener/pull/9624. There is https://github.com/gardener/gardener/pull/9802 to fix it.

TL;DR: This PR is unrelated to this g/g issue.

charts/internal/calico-monitoring/templates/configmap-dashboards.yaml
```diff
@@ -0,0 +1,15 @@
...
+  name: calico-dashboards
+  namespace: {{ .Release.Namespace }}
+  labels:
+    dashboard.monitoring.gardener.cloud/shoot: "true"
+data:
+  calico-felix-dashboard.json: |-
+  {{- .Files.Get "calico-felix-dashboard.json" | nindent 4 }}
+
+  calico-typha-dashboard.json: |-
+  {{- .Files.Get "calico-typha-dashboard.json" | nindent 4 }}
```
ScheererJ, line 14: Please replace datasource `-- Grafana --` with `-- Plutono --` in both dashboard files.

```
-        "datasource": "-- Grafana --",
+        "datasource": "-- Plutono --",
```

Otherwise there will be an ugly error shown more or less randomly complaining about the unknown datasource.

ScheererJ (changes requested):
Thanks for adapting the extension to GEP-19. Please update the dashboard files so that nobody gets ugly error messages.

ScheererJ (approved):
/lgtm

PR #437, opened 2024-07-08
[admission] Add `priorityClassName`

timuthy:

/area robustness
/area networking
/kind enhancement
/platform azure

This PR enhances the admission chart by the option to set a `priorityClassName`.

Part of https://github.com/gardener/gardener/issues/9936.

**Release note**:

```other operator
A `priorityClassName` can now be set for the admission deployment via the `gardener-extension-admission-calico` Helm chart.
```

### DIFF
--- a/charts/gardener-extension-admission-calico/values.yaml
+++ b/charts/gardener-extension-admission-calico/values.yaml
@@ -7,6 +7,7 @@ global:
     repository: europe-docker.pkg.dev/gardener-project/public/gardener/extensions/admission-calico
     tag: latest
     pullPolicy: IfNotPresent
+# priorityClassName: gardener-garden-system-400
   replicaCount: 1
   resources: {}
   healthPort: 8081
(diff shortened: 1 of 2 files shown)


axel7born:
/lgtm
Just wondering about `/platform azure` in the description?

axel7born (approved):

charts/gardener-extension-admission-calico/values.yaml
```diff
@@ -7,6 +7,7 @@ global:
     repository: europe-docker.pkg.dev/gardener-project/public/gardener/extensions/admission-calico
     tag: latest
     pullPolicy: IfNotPresent
+# priorityClassName: gardener-garden-system-400
```
ialidzhikov, line 10: Any reason not to enable it by default but to require the component that installs the admission charts to explicitly pass it?
If there is no concrete reason, I would suggest to set it by default. Motivation: to have more sane defaults that work out of the box.

PR #464, opened 2024-08-13
Add support for static resource requests

ScheererJ:

/area networking
/area auto-scaling
/kind enhancement

Add support for static resource requests.

While it is generally best practice to utilise automatic resource management and scaling in kubernetes environments there may be situations when the workload is static or any disruptions may cause serious issues. In such cases it is better to be able to use static resource allocation instead of dynamic autoscaling like cluster-proportional or vertical pod autoscaler. This is in particular true for system components such as calico, which may cause issues in the pod network if a process is not available for an extended amount of time.

Fixes #

**Release note**:

```other operator
Add static resource allocation autoscaling mode for calico node/typha (autoScaling.mode: static).
```


pkg/apis/calico/types_network.go
```diff
@@ -155,10 +156,22 @@ const (
 	AutoscalingModeClusterProportional AutoscalingMode = "cluster-proportional"
 	// AutoscalingModeVPA is a constant for vertical pod autoscaling mode.
 	AutoscalingModeVPA AutoscalingMode = "vpa"
+	// AutoscalingModeStatic is a constant for static resource allocation as autoscaling mode.
+	AutoscalingModeStatic AutoscalingMode = "static"
 )
 
 // AutoScaling defines how the calico components are automatically scaled. It allows to use either vertical pod or cluster-proportional autoscaler (default: cluster-proportional).
 type AutoScaling struct {
 	// Mode defines how the calico components are automatically scaled. It allows to use either vertical pod or cluster-proportional autoscaler (default: cluster-proportional).
```
axel7born, line 165: `static` mode should be added to this comment.

docs/usage/usage.md
```diff
@@ -81,6 +80,25 @@ autoScaling:
...
+  resources:
+    node:
+      cpu: 100m
+      memory: 100Mi
+    typha:
+      cpu: 100m
+      memory: 100Mi
+```
+
+> ℹ️ Please note that it is only possible to (optionally) configure the resource requests for `calico-node` and `calico-typha` in static mode.
```
axel7born, line 99: I'm not sure if this sentence is really understandable. At least, I got it wrong reading it for the first time.
It is not needed, to configure both `calico-node` **and** `calico-typha` and the `optionally` means that if resources are not specified there is a default.
Maybe, we should use the default values in this example.
suggestion:
```
"Please note that in static mode, you have the option to configure the resource requests for calico-node and calico-typha. If not specified, default settings will be used."
```

axel7born:
The PR looks good to me. I have only small suggestions regarding docs or comments in the code.

axel7born (changes requested):

docs/usage/usage.md
```diff
@@ -65,14 +65,13 @@ spec:
 
 ## AutoScaling
 
-Autoscaling defines how the calico components are automatically scaled. It allows to use either vertical pod or cluster-proportional autoscaler (default: cluster-proportional).
+Autoscaling defines how the calico components are automatically scaled. It allows to use either static resource assignment, vertical pod or cluster-proportional autoscaler (default: cluster-proportional).
 
-The cluster-proportional autoscaling mode is preferable when conditions require minimimal disturbances and vpa mode for improved cluster resource utilization. 
+The cluster-proportional autoscaling mode is preferable when conditions require minimal disturbances and vpa mode for improved cluster resource utilization. Static resource assignments causes the least disruptions, but has no dynamics to handle changing demands.
```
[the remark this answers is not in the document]
ScheererJ: Thanks for the simplification. Added with https://github.com/gardener/gardener-extension-networking-calico/pull/464/commits/d66173fd499a37fa5a80f927d5fe0cfa0a719f2d.

pkg/apis/calico/types_network.go
ScheererJ: Adapted the documentation strings of corresponding fields/types in https://github.com/gardener/gardener-extension-networking-calico/pull/464/commits/d66173fd499a37fa5a80f927d5fe0cfa0a719f2d.

docs/usage/usage.md
ScheererJ: Adapted the section to hopefully be more clear.
WDYT?

axel7born (approved):

PR #486, opened 2024-09-11
Fix networkConfig for IPv6.

axel7born:
`networkConfig` should be there, even if `Networking.Nodes` is not configured.

/area networking
/kind bug

Fixes #

**Release note**:

```other operator
Fix networkConfig for IPv6.
```


ScheererJ (approved):
/lgtm

PR #489, opened 2024-09-19
Adjust resource requests and autoscaling of `calico-node` side car containers.

ScheererJ:

/area networking
/kind enhancement

Adjust resource requests and autoscaling of `calico-node` side car containers.

The side car containers use negligible amount of resources compared to `calico-node`. Therefore, requests likely just waste resources. Similarly, they fall below the minimum threshold of VPA. Hence, autoscaling them does not make sense.
`minAllowed` and `maxAllowed` of `calico-node` also are not that useful. Therefore, they are removed as well.

Fixes #

**Release note**:

```other operator
In VPA autoscaling mode, `calico-node` should be disrupted less often as side car containers are no longer considered by VPA. Additionally, the minimum/maximum restriction are removed, which can lead to less memory consumption.
```

### DIFF
--- a/charts/internal/calico/templates/node/daemonset-calico-node.yaml
+++ b/charts/internal/calico/templates/node/daemonset-calico-node.yaml
@@ -171,10 +171,6 @@ spec:
           env:
           - name: POD_CIDR
             value: {{ .Values.global.podCIDR }}
-          resources:
-            requests:
-              cpu: 10m
-              memory: 50Mi
           command:
           - /bin/sh
           - -c
@@ -238,10 +234,6 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
-          resources:
-            requests:
-              cpu: 1m
-              memory: 20Mi
           command:
             - /bin/sh
             - -c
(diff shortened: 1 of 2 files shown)


axel7born (approved):

vlerenc:
Hmm... isn't removing a container from the VPA spec resulting in applying the default configuration, i.e. whatever is defined in `spec.updatePolicy`? That would keep all "removed" containers in `updateMode` `Auto` (instead of `off`) and even set `controlledValues` to `RequestsAndLimits` (even worse than `RequestsOnly`)?

cc @voelzmo

ScheererJ:
> Hmm... isn't removing a container from the VPA spec resulting in applying the default configuration, i.e. whatever is defined in `spec.updatePolicy`? That would keep all "removed" containers in `updateMode` `Auto` (instead of `off`) and even set `controlledValues` to `RequestsAndLimits` (even worse than `RequestsOnly`)?
> 
> cc @voelzmo

Good point. Adjusted accordingly with https://github.com/gardener/gardener-extension-networking-calico/pull/489/commits/0f5afa276c4f49aa43904ae38b73134b11bca214.

axel7born (approved):

vlerenc:
Thanks @ScheererJ. I also checked in the meantime VPA and it requires the container indeed to be put explicitly into `off` mode if the general policy is `auto`.

PR #491, opened 2024-09-23
Remove VPA `(min|max)Allowed` settings.

ScheererJ:

/area networking
/area auto-scaling
/kind enhancement

Remove VPA `(min|max)Allowed` settings.

`minAllowed` may cause wasted resources while `maxAllowed` may cause under-reservation and hence noisy neighbor issues. This change removes them building on previous work in #489. They are removed from `kube-controllers`, `calico-typha`, the admission controller and the extension.

Fixes #

**Release note**:

```other operator
The networking calico extension no longer configures min/maxAllowed in any managed VPA resource.
```


charts/gardener-extension-admission-calico/values.yaml
```diff
@@ -13,12 +13,12 @@ global:
...
-        memory: 64Mi
-      maxAllowed:
-        cpu: 4
-        memory: 10G
+    resourcePolicy: {}
+#     minAllowed:
+#       memory: 64Mi
+#     maxAllowed:
+#       cpu: 4
+#       memory: 10G
```
axel7born, line 21: Why is this commented and not removed?

DockToFuture (approved):
/lgtm

charts/gardener-extension-admission-calico/values.yaml
ScheererJ: It is still possible to set it by specifying it. The comments make it easy to see how to set it when deploying the extension/admission controller.

axel7born (approved):

PR #506, opened 2024-10-17
set correct iptables rules when using ipv6

DockToFuture:

/area networking
/kind enhancement

Correct iptable backend and iptable rule are set for IPv6 shoot clusters when running with node-local-dns.

Fixes #

**Release note**:

```other operator
Correct iptable backend and iptable rule are set for IPv6 shoot clusters when running with node-local-dns.
```


ScheererJ (approved):
/lgtm

PR #521, opened 2024-11-06
Sanity Check for dual-stack before use IPAM.CIDR 

mohamed-rafraf:

This PR introduces a boolean check to determine if the cluster is dual-stack. This check prevents misconfiguration when the user specifies an `ipam.cidr` value. If `ipam.cidr` is set , it will lead to an erroneous Calico configuration for dual-stack clusters.

Example configuration:

```yaml
networking:
  ipFamilies:
    - IPv4
    - IPv6
  type: calico
  nodes: 10.250.0.0/16
  providerConfig:
    apiVersion: calico.networking.extensions.gardener.cloud/v1alpha1
    kind: NetworkConfig
    ipam:
      type: host-local
      cidr: usePodCIDR
```

In this setup, specifying `ipam.cidr` can result in the Calico configuration containing incorrect `ranges` and `subnet` fields (an invalid configuration for dual-stack networks).

/area networking
/kind bug

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/charts/utils.go
+++ b/pkg/charts/utils.go
@@ -311,7 +311,7 @@ func mergeCalicoValuesWithConfig(c *calicoConfig, config *calicov1alpha1.Network
 	}
 
 	if c.IPAM.IPAMType == hostLocal {
-		if config.IPAM != nil && config.IPAM.CIDR != nil {
+		if config.IPAM != nil && config.IPAM.CIDR != nil && !(isIPv4 && isIPv6) {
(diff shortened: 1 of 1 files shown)


gardener-robot-ci-1:
Thank you @mohamed-rafraf for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

DockToFuture (approved):
/lgtm

PR #563, opened 2025-01-16
do not enable ipinip for non-overlay case

DockToFuture:

/area networking
/kind bug

Do not enable IPIP for non-overlay case when no networkConfig is set.
Fixes #

**Release note**:

```other operator
Do not enable IPIP for non-overlay case when no networkConfig is set.
```


pkg/controller/actuator_reconcile.go
```diff
@@ -144,8 +144,6 @@ func (a *actuator) Reconcile(ctx context.Context, _ logr.Logger, network *extens
 		} else {
 			if ipFamilies.Has(extensionsv1alpha1.IPFamilyIPv6) {
```
axel7born, line 145: Maybe, this `if` should also be removed, resulting in `setPoolMode(networkConfig, ipFamilies, calicov1alpha1.Never) for both IP-Families?

ScheererJ (approved):
/lgtm

PR #656, opened 2025-05-15
Improved signal handling on termination

toanju:
The daemonset calico-node does not terminate gracefully on SIGTERM.
Therefor the pod is lingering until it gets killed after a timeout. The
proposed patch switches to a `sleep infinity` and adds a `wait` which is
returning due to the added `trap`. This will terminate the pod in due
time.


gardener-robot-ci-1:
Thank you @toanju for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

ScheererJ (approved):
Thanks for the improvement.

/lgtm

PR #687, opened 2025-07-24
Bird exporter as calico-node sidecar

videlov:
/area networking
/area monitoring
/kind enhancement

Introduce a new sidecar container to `calico-node` DaemonSet exporting metrics from `/var/run/calico/bird.ctl` socket. Also include needed Service + ScrapeConfig for collecting metrics into Shoot Prometheus.

Fixes https://github.com/ironcore-dev/gardener-extension-provider-ironcore-metal/issues/213

**Release note**:
```other operator
`bird-exporter` sidecar added to `calico-node` DaemonSet to export `bird` metrics into Prometheus
```


charts/internal/calico-monitoring/templates/scrapeconfig-bird.yaml
```diff
@@ -0,0 +1,65 @@
...
+      - kube-system
+    role: Endpoints
+    tlsConfig:
+      # This is needed because we do not fetch the correct cluster CA bundle right now
+      insecureSkipVerify: true
+  metricRelabelings:
+  - sourceLabels:
+    - __name__
+    action: keep
+    regex: bird_.+
```
rickardsjp, line 36: This retains all metrics that aren't the "meta" metrics. Do you need all of them?

rickardsjp (commented):
Thank you for your PR @videlov!

From a monitoring perspective the scrape configs look fine. However, I have some concerns about the cardinality of metrics and the impact on retention. From what I can see the metrics are exposed for each node about each of its BGP neighbors. This means the number of metrics scraped and ingested grows in a quadratic fashion. When this is enabled on clusters with hundreds of nodes, this means we can have hundreds of thousands of additional time series that are ingested.
Metric cardinality doesn't influence metric retention proportionally, because Prometheus performs optimization when storing these metrics. However, the Gardener shoot control plane Prometheus does not scale its disk and retention with the cluster size (number of nodes) and I worry that this will negatively impact retention.

Is it intentional not to include a dashboard? Felix and Typha both have dashboards, and I saw that the `bird_exporter` repository has a [dashboard](https://github.com/czerwonk/bird_exporter/blob/main/grafana/BIRD-RS-1522082867844.json) as well. How do you intend to use the metrics?

Lastly, I had a quick offline discussion with @ScheererJ and @axel7born about this PR and we noted that this PR puts the enablement of the bird exporter in the hands of cluster owners. Bird exporter metrics are then opt-in for the cluster owner. Would it make more sense to make this configurable on the extension so that it can be enabled on a landscape level instead?

rickardsjp:
One more point, the linked issue assesses the original bird exporter project to be "quite unmaintained" to the point where SAP is maintaining a fork of the project. It is not evident from the issue or this PR why you're adding the original project to this PR. Please explain why.

charts/internal/calico-monitoring/templates/scrapeconfig-bird.yaml
defo89: We are mostly interested in `bird_protocol_up` to monitor BGP session state. 

Additionally, `bird_protocol_prefix_.` metrics might be interesting. Other protocols like OSPF are not relevant for for Calico CNI use case. We could narrow them to above and extend if required.

charts/internal/calico/templates/node/daemonset-calico-node.yaml
```diff
@@ -498,6 +498,30 @@ spec:
             - name: cni-log-dir
               mountPath: /var/log/calico/cni
               readOnly: true
+        {{- if and .Values.config.birdExporter.enabled (not .Values.config.nonPrivileged) }}
```
ScheererJ, line 501: I guess we also need to check for `.Values.config.monitoring.enabled` as the port is defined there similarly to the check in `[bird-monitoring-service.yaml](https://github.com/gardener/gardener-extension-networking-calico/pull/687/files#diff-e235f325163ebd134ab429f6eef255a49b4e09268aca62f49efe15499722bfdf)`.

ScheererJ (commented):
Thanks for the interesting enhancement.

Do you think it makes sense to also add a dashboard for the new metrics similar to the felix/typha dashboards?

charts/internal/calico-monitoring/templates/scrapeconfig-bird.yaml
videlov: Makes sense, we did evaluate and did reduce them. Basically we need eventually only `bird_protocol_up` for our alerts, but will be also keeping `bird_protocol_uptime`, `bird_protocol_prefix_export_count` and `bird_protocol_prefix_export_count`.

PR #719, opened 2025-09-23
add Multus CNI

defo89:

/area networking
/kind enhancement

Add support for [multus-cni](https://github.com/k8snetworkplumbingwg/multus-cni) to enable attaching additional network interfaces to pods. Also add an option to install [containernetworking/plugins](https://github.com/containernetworking/plugins) on the node.

- Deployments [example](https://github.com/k8snetworkplumbingwg/multus-cni/tree/master/deployments) and [mellanox deployment](https://github.com/Mellanox/network-operator/blob/master/manifests/state-multus-cni/0050-multus-ds.yml) were used as a baseline for this PR.

- For convenience reasons additional common CNI plugins are fetched from [k8snetworkplumbingwg/plugins](https://github.com/k8snetworkplumbingwg/plugins/pkgs/container/plugins). We can switch to upstream when https://github.com/containernetworking/plugins/pull/844 is merged.

**Release note**:

```other operator
Add `multus-cni` DaemonSet to enable attaching multiple network interfaces to pods
```


charts/internal/calico/templates/multus/daemonset-multus-cni.yaml
```diff
@@ -0,0 +1,136 @@
...
+  name: multus
+  namespace: kube-system
+  labels:
+    k8s-app: multus
+    gardener.cloud/role: system-component
+    node.gardener.cloud/critical-component: "true"
+  {{- if not .Values.autoscaling.staticRequests }}
+  annotations:
+    resources.gardener.cloud/preserve-resources: 'true'
+  {{- end }}
```
ScheererJ, line 17: I think we do not need this unless we autoscale the daemonset, which does not seem to be the case right now.

What does this daemonset actually do at runtime besides installing the CNI binaries with the init container?

charts/internal/calico/templates/multus/daemonset-multus-cni.yaml
```diff
@@ -0,0 +1,136 @@
...
+            - --readiness-indicator-file=/etc/cni/net.d/10-calico.conflist
+          securityContext:
+            privileged: true
+          resources:
+            requests:
+              cpu: "100m"
+              memory: "200Mi"
+            limits:
+              cpu: "100m"
+              memory: "200Mi"
```
ScheererJ, line 118: As far as I can tell, this container does some initial setup, but afterwards it just does some polling every second that the CNI configuration did not change (see https://github.com/k8snetworkplumbingwg/multus-cni/blob/00adf22482945de41ff702258593aaf867a48e2d/cmd/thin_entrypoint/main.go#L627-L663). For that the resource requests seem a little high. WDYT?

In general, we try to stay away from CPU limits, i.e. we removed them almost everywhere in the gardener code base, because they have potential for a lot of issues, but help rarely. Do you think that we really need them here?

charts/internal/calico/templates/multus/daemonset-multus-cni.yaml
```diff
@@ -0,0 +1,136 @@
...
+    type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 2
+  template:
+    metadata:
+      labels:
+        node.gardener.cloud/critical-component: "true"
+        networking.gardener.cloud/to-public-networks: allowed
+        networking.gardener.cloud/to-apiserver: allowed
+        networking.gardener.cloud/to-dns: allowed
```
ScheererJ, line 33: Is this daemonset really talking to the world? I can completely understand the API server and DNS access, but public networks seems surprising. Could you please explain?

charts/internal/calico/templates/multus/daemonset-multus-cni.yaml
```diff
@@ -0,0 +1,136 @@
...
+        k8s-app: multus
+        gardener.cloud/role: system-component
+    spec:
+      nodeSelector:
+        kubernetes.io/os: linux
+      hostNetwork: true
+      hostPID: true
```
ScheererJ, line 40: Is this required? I find it surprising.

charts/internal/calico/templates/multus/daemonset-multus-cni.yaml
```diff
@@ -0,0 +1,136 @@
...
+          - "--type"
+          - "thin"
+        resources:
+          requests:
+            cpu: 10m
+            memory: 15Mi
+        volumeMounts:
+          - name: cnibin
+            mountPath: /host/opt/cni/bin
+            mountPropagation: Bidirectional
```
ScheererJ, line 71: Does this volume mount need to be `Bidirectional`?

charts/internal/calico/templates/multus/daemonset-multus-cni.yaml
```diff
@@ -0,0 +1,136 @@
...
+        securityContext:
+          privileged: true
+      {{ if .Values.config.multus.installCNIPlugins }}
+      # This container installs the CNI binaries on each node.
+      - name: install-cni-plugins
```
ScheererJ, line 76: What does this container install? I thought that `multi-install-cni` container would install multus. What is copied here?

charts/internal/calico/templates/multus/daemonset-multus-cni.yaml
```diff
@@ -0,0 +1,136 @@
...
+          args:
+            - --cleanup-config-on-exit
+            - --cni-bin-dir=/host/opt/cni/bin
+            - --cni-conf-dir=/host/etc/cni/net.d
+            - --cni-version=0.3.1
+            - --multus-autoconfig-dir=/host/etc/cni/net.d
+            - --multus-cni-conf-dir=/etc/cni/multus/net.d
+            - --multus-conf-file=auto
+            - --multus-log-file=/var/log/multus.log
+            - --multus-log-level=error
```
ScheererJ, line 106: Is the log level sufficient? I am wondering whether we need more than just errors. However, we do not want the logs to be spammed. I am not sure which log levels are available here and what the recommended option would be. Could you please elaborate?

charts/internal/calico/values.yaml
ScheererJ, line 57: Please add the two multus images here.

ScheererJ (commented):
Thanks for the enhancement. It opens up new scenarios.

charts/internal/calico/templates/multus/daemonset-multus-cni.yaml
defo89: I think I have copied this bit over from `calico-node` DaemonSet.

Agree, it's not doing much. Though the running the main container with `thin_entrypoint` is also important that it enters the watch (based on conditions https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/cmd/thin_entrypoint/main.go#L621-L634 ) and updates the multus kubeconfig:

```shell
› kubectl-v1.31.5 -n kube-system logs multus-g4gzj
Defaulted container "multus" out of: multus, install-multus-cni (init), install-cni-plugins (init)
CA (false) or SA token (true) changed - recreating kubeconfig
kubeconfig is created in /host/etc/cni/net.d/multus.d/multus.kubeconfig
```

charts/internal/calico/templates/multus/daemonset-multus-cni.yaml
defo89: `hostPID: true` is a `thick` mode testing leftover - removed

charts/internal/calico/templates/multus/daemonset-multus-cni.yaml
```diff
@@ -0,0 +1,130 @@
+{{- if .Values.config.multus.enabled }}
+---
+# This manifest installs Multus CNI plugins and network config on
+# each node in a Kubernetes cluster.
+kind: DaemonSet
```
axel7born, line 5: What is the reason for add an additional `DaemonSet` and not adding the container to `calico-node`?
Wouldn't it make sense to have an own extensions for multus?
Couldn't is work for Cilium too?

charts/internal/calico/templates/multus/crd-networkattachmentdefinition.yaml
```diff
@@ -0,0 +1,47 @@
...
+      schema:
+        openAPIV3Schema:
+          description: 'NetworkAttachmentDefinition is a CRD schema specified by the Network Plumbing
+            Working Group to express the intent for attaching pods to one or more logical or physical
+            networks. More information available at: https://github.com/k8snetworkplumbingwg/multi-net-spec'
+          type: object
+          properties:
+            apiVersion:
+              description: 'APIVersion defines the versioned schema of this represen
+                tation of an object. Servers should convert recognized schemas to the
```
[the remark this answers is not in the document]
defo89: Fixed, that was a copy paste from upstream https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/deployments/multus-daemonset.yml#L35-L37

charts/internal/calico/templates/multus/daemonset-multus-cni.yaml
```diff
@@ -0,0 +1,130 @@
...
+      {{ if .Values.config.multus.installCNIPlugins }}
+      # This container installs the CNI binaries on each node.
+      - name: install-cni-plugins
+        image: {{ index .Values.images "cni-plugins" }}
+        command: ["/bin/sh"]
+      # Do not install host-local, portmap, loopback and tuning plugins as they are already provided by Calico
+      # see https://github.com/projectcalico/calico/blob/dbf63126b7297bfd8a401d6bef61d640fa8f250f/cni-plugin/Makefile#L147
+        args:
+          - "-c"
+          - "cd /usr/src/cni/bin && cp $(ls | grep -v -E '^(host-local|portmap|loopback|tuning)$') /host/opt/cni/bin/"
```
DockToFuture, line 76: I would have preferred a direct find and then doing the copy:
```
find /usr/src/cni/bin -maxdepth 1 -type f ! -name 'host-local' ! -name 'portmap' ! -name 'loopback' ! -name 'tuning' -exec cp {} /host/opt/cni/bin/ \;
```

charts/internal/calico/templates/multus/daemonset-multus-cni.yaml
defo89: I saw possible cons of adding additional containers to `calico-node`:
- Longer startup time for calico-node
- While adding init containers that copy over the binaries would be fine to add, we also need an actively running main container with `thin_entrypoint` that updates multus kubeconfig.
- Theoretically (not sure about practicality of this) if multus does not start up for whatever reason and `/etc/cni/net.d/00-multus.conf` is not written, we have a fully functioning cluster with CNI

In the initial discussion with @ScheererJ we thought that it would be an overkill to create a separate extension for this use case (development and maintenance overhead).

Additionally, the multus config is "tied" to a specific CNI and its config. This part is important:
```
            - --multus-master-cni-file-name=10-calico.conflist
            - --readiness-indicator-file=/etc/cni/net.d/10-calico.conflist
```
Maybe it can be computed based on configured CNI but adds more complexity.

wdyt?

axel7born: Thanks for the explanation. That makes sense to me.
If we really need multus for Cilium, we can think about an own extension.

charts/internal/calico/templates/multus/daemonset-multus-cni.yaml
```diff
@@ -0,0 +1,130 @@
...
+      containers:
+        # Runs multus container on each Kubernetes node.
+        - name: multus
+          image: {{ index .Values.images "multus-cni" }}
+          command: ["/thin_entrypoint"]
+          args:
+            - --cleanup-config-on-exit
+            - --cni-bin-dir=/host/opt/cni/bin
+            - --cni-conf-dir=/host/etc/cni/net.d
+            - --cni-version=0.3.1
```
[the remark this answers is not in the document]
defo89: Sounds good to me. Do you think we should expose this externally via API or just have it set centrally in `values.yaml`? From my POV latter should be enough.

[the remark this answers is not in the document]
defo89: added `cniVersion ` for now

DockToFuture (approved):
/lgtm

PR #732, opened 2025-10-17
Use separate port for healthz endpoint if running in ebpf mode with kube-proxy

ScheererJ:

/area networking
/kind enhancement

Use separate port for healthz endpoint if running in ebpf mode with kube-proxy.

In ebpf mode `calico-node` has to expose the healthz endpoint for load balancers if it takes over service routing, i.e. if `kube-proxy` is not running. However, it should not clash with `kube-proxy` if running in ebpf mode together with `kube-proxy`. Therefore, we set the healthz endpoint to a different port in this case.

Fixes #

Setting the environment with typo as with the actual change (see https://github.com/projectcalico/calico/issues/10748) and with the proper spelling in case it gets corrected.

**Release note**:

```other operator
`calico-node` should not longer bind to the `kube-proxy` healthz port if used in ebpf mode and `kube-proxy` is enabled.
```

### DIFF
--- a/charts/internal/calico/templates/node/daemonset-calico-node.yaml
+++ b/charts/internal/calico/templates/node/daemonset-calico-node.yaml
@@ -383,6 +383,13 @@ spec:
             # Enable eBPF dataplane mode.
             - name: FELIX_BPFENABLED
               value: "{{ .Values.config.felix.bpf.enabled }}"
+            # Move Kube-proxy healthz endpoint of calico-node to a different port when running in eBPF mode and kube-proxy is enabled.
+            {{- if and .Values.config.felix.bpf.enabled (not .Values.config.felix.bpfKubeProxyIPTablesCleanup.enabled) }}
+            - name: FELIX_BPFKUBEPROXYHEALTZPORT
+              value: "10257"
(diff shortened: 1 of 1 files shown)


charts/internal/calico/templates/node/daemonset-calico-node.yaml
```diff
@@ -383,6 +383,13 @@ spec:
             # Enable eBPF dataplane mode.
             - name: FELIX_BPFENABLED
               value: "{{ .Values.config.felix.bpf.enabled }}"
+            # Move Kube-proxy healthz endpoint of calico-node to a different port when running in eBPF mode and kube-proxy is enabled.
+            {{- if and .Values.config.felix.bpf.enabled (not .Values.config.felix.bpfKubeProxyIPTablesCleanup.enabled) }}
+            - name: FELIX_BPFKUBEPROXYHEALTZPORT
```
domdom82: can we add a TODO here to remove this once the typo is fixed upstream? maybe add a link to the typo.

otherwise we lose context from the PR later and only see the two lines which may be confusing.

domdom82: I will add it in a follow-up PR

PR #733, opened 2025-10-20
disable ipinip if overlay is set

DockToFuture:

/area networking
/kind bug

Ensure `FELIX_IPINIPENABLED` is set to false in all cases if overlay is disabled. 

Fixes #

**Release note**:

```other operator
Ensure `FELIX_IPINIPENABLED` is set to false in all cases if overlay is disabled. 
```

### DIFF
--- a/pkg/charts/utils.go
+++ b/pkg/charts/utils.go
@@ -292,6 +292,11 @@ func generateChartValues(network *extensionsv1alpha1.Network, config *calicov1al
 		)
 	}
 
+	// Disable IP-in-IP if overlay is disabled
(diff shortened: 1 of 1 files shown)


domdom82:
@DockToFuture just curious to see if my understanding is correct:
> This leads to problems when disabling the overlay as the FelixConfiguration takes precedence over environment variables

however, the linked [documentation](https://docs.tigera.io/calico/latest/reference/felix/configuration) says
> The value of any configuration parameter is the value read from the first location containing a value. For example, if an environment variable contains a value, it takes top precedence.

To me this reads like the opposite of your claim. Am I missing something?

PR #741, opened 2025-10-31
feat: log dropped packets due to MTU issues

domdom82:

/area networking
/kind enhancement

- We've seen several connectivity issues with overlay clusters and VPN in the past.
- We were able to rca those issues to wrong MTU/MSS values set on tcp connections forwarded by VPN
- The issues got fixed, however this PR will add more observability in case similar issues arise in the future
- The PR installs an iptables logging rule to calico clusters that prints a `PACKET-TOO-BIG` line to the kernel log in case a `fragentation-needed` (v4) or `packet-too-big` (v6) packets was issued by the local host due to bad MTU settings.
- The PACKET-TOO-BIG string can be easily searched for in plutono/vali if needed.

**Release note**:

```other operator
Calico clusters will now report MTU issues to the kernel log using a PACKET-TOO-BIG prefix.
```

### DIFF
--- a/charts/internal/calico/templates/node/daemonset-calico-node.yaml
+++ b/charts/internal/calico/templates/node/daemonset-calico-node.yaml
@@ -61,6 +61,35 @@ spec:
         securityContext:
           privileged: true
       {{- end }}
+      # Log MTU issues if monitoring is enabled
+      {{- if .Values.config.monitoring.enabled }}
+      - name: log-mtu-issues
+        image: {{ index .Values.images "calico-node" }}
+        command:
+          - /bin/sh
+          - -c
+          - |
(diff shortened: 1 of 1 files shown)


charts/internal/calico/templates/node/daemonset-calico-node.yaml
```diff
@@ -61,6 +61,34 @@ spec:
...
+        image: {{ index .Values.images "calico-node" }}
+        command:
+          - /bin/sh
+          - -c
+          - |
+            if [ -x "/usr/sbin/iptables-legacy" ]; then
+              backend="legacy"
+            else
+              backend="nft"
+            fi
```
axel7born: For this rules it might not  be important. But for consistency we should decide between legacy and nft based on the kublet rules like  [here](https://github.com/gardener/gardener-extension-networking-calico/blob/7cc6703866cba7f8892b8df22a97a969ca05a8ab/charts/internal/calico/templates/node/daemonset-calico-node.yaml#L180-L185)

domdom82: thanks, I've tested it and it works fine. updated the PR accordingly.

PR #843, opened 2026-05-19
Disable IP-in-IP when overlay is switched off

axel7born:

/area networking
/kind bug

When `spec.overlay.enabled: false` is set, the Calico IP-in-IP encapsulation serves no purpose — the cluster is intended to route pods directly. However, Felix was still creating a `tunl0` device and factoring tunnel overhead into MTU calculations, which wasted resources and could cause subtle networking issues.

This fix explicitly sets `Felix.IPInIP.Enabled = false` and `IPv4.Mode = Never` whenever overlay is disabled.
Fixes #

**Release note**:

```other operator
Disable Felix IP-in-IP when overlay is explicitly disabled to prevent unnecessary tunnel device creation and incorrect MTU overhead.
```

### DIFF
--- a/pkg/charts/utils.go
+++ b/pkg/charts/utils.go
@@ -353,6 +353,13 @@ func mergeCalicoValuesWithConfig(c *calicoConfig, config *calicov1alpha1.Network
 		c.IPv4.Mode = calicov1alpha1.Never
 	}
 
+	// When overlay is explicitly disabled, IPIP serves no purpose. Disable it so Felix does
+	// not create a tunl0 device and does not subtract tunnel overhead from the MTU.
+	if config.Overlay != nil && !config.Overlay.Enabled {
+		c.Felix.IPInIP.Enabled = false
+		c.IPv4.Mode = calicov1alpha1.Never
+	}
+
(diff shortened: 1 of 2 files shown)


pkg/charts/utils.go
```diff
@@ -353,6 +353,13 @@ func mergeCalicoValuesWithConfig(c *calicoConfig, config *calicov1alpha1.Network
 		c.IPv4.Mode = calicov1alpha1.Never
 	}
 
+	// When overlay is explicitly disabled, IPIP serves no purpose. Disable it so Felix does
+	// not create a tunl0 device and does not subtract tunnel overhead from the MTU.
+	if config.Overlay != nil && !config.Overlay.Enabled {
```
domdom82, line 358: what if `config.Overlay` is not set? do we still default to IP-in-IP for Felix?

Looks to me [like we do](https://github.com/gardener/gardener-extension-networking-calico/blob/master/pkg/charts/utils.go#L128)?

axel7born: Yes, this is consistent with the default, that overlay is enabled when no overlay config is present.

PR #916, opened 2026-09-07
Omit `shoot-calico-typha` scape config from deployed chart resources if disabled

LucaBernstein:

/area networking
/kind bug

If calico-typha is disabled, its scrape configuration should also not be deployed.

https://github.com/gardener/gardener/pull/15662

/cc @ScheererJ 

**Release note**:

```bugfix operator
Skip deployment of `shoot-calico-typha` `ScrapeConfig` if Typha is disabled.
```


pkg/controller/actuator_reconcile.go
```diff
@@ -79,31 +79,42 @@ func applyMonitoringConfig(ctx context.Context, seedClient client.Client, chartA
...
 
+	typhaEnabled := true
 	if network.Spec.ProviderConfig != nil && network.Spec.ProviderConfig.Raw != nil {
 		networkConfig, err := calicov1alpha1helper.CalicoNetworkConfigFromNetworkResource(network)
 		if err != nil {
 			return err
 		}
+		if networkConfig.Typha != nil && !networkConfig.Typha.Enabled {
+			typhaEnabled = false
+		}
```
ScheererJ:
```suggestion
		if networkConfig.Typha != nil {
			typhaEnabled = networkConfig.Typha.Enabled
		}
```

pkg/controller/actuator_reconcile.go
```diff
@@ -79,31 +79,42 @@ func applyMonitoringConfig(ctx context.Context, seedClient client.Client, chartA
...
 		return client.IgnoreNotFound(calicoControlPlaneMonitoringChart.Delete(ctx, seedClient, network.Namespace))
 	}
 
-	return calicoControlPlaneMonitoringChart.Apply(ctx, chartApplier, network.Namespace, nil, "", "", nil)
+	chartValues := map[string]any{
+		"typha": map[string]any{
+			"enabled": typhaEnabled,
+		},
+	}
+	return calicoControlPlaneMonitoringChart.Apply(ctx, chartApplier, network.Namespace, nil, "", "", chartValues)
```
ScheererJ: Why is the scrap config enabled twice?
1. Once by adding it to the monitoring chart objects.
2. Once by enabling via the chart values.

pkg/controller/actuator_reconcile.go
ScheererJ, line 1: How is deletion handled? It looks like this only covers creation of the resources, but how is the scrape configuration deleted if I decide to disable typha in an existing cluster?

ScheererJ (commented):
Thanks for addressing the gap in the monitoring resource handling.

LucaBernstein:
Thank you @ScheererJ for your review!
I made a few enhancements and changes:
- Both `ScrapeConfig`s are now always kept in the `Objects` which means that they are made known to the chart framework. This is (only) relevant for the `.Delete()` path of the chart framework.
- If Typha (or Bird) is disabled after being enabled previously, their `ScrapeConfig` removal will now be handled by added cleanup logic that explicitly deletes the resources in the cluster.
- This also resolves your question about the configs being presumably added twice - `Objects` ist just about registering to the chart framework (for deletion). :)

pkg/controller/actuator_reconcile.go
```diff
@@ -83,27 +83,50 @@ func applyMonitoringConfig(ctx context.Context, seedClient client.Client, chartA
...
+	disabledScrapeConfigs := map[string]bool{
+		"shoot-calico-typha": !typhaEnabled,
+		"shoot-calico-bird":  !birdEnabled,
+	}
+	for _, o := range calicoControlPlaneMonitoringChart.Objects {
+		if disabledScrapeConfigs[o.Name] {
+			if err := o.Delete(ctx, seedClient, network.Namespace); err != nil {
+				return err
+			}
+		}
```
ScheererJ, line 126: Do we not need to ignore not found errors here?

LucaBernstein: If you take a look at the implementation, it contains:
```go
	if err := c.Get(ctx, key, obj); err != nil {
		if apierrors.IsNotFound(err) {
			return nil
		}
		return fmt.Errorf("could not get %T '%s': %w", obj, key.String(), err)
	}
```
so I deemed this already covered.

LucaBernstein: Added.

ScheererJ (approved):
/lgtm
/approve

PR #920, opened 2026-09-14
Restart Typha after CPM and HA transition to fix stale watch cache

axel7born:

/area networking
/kind bug

During control plane migration and non-HA -> HA transitions, calico-typha
can reconnect to the new kube-apiserver while its watch cache for Calico CRD
types is still warming up. This causes Typha to serve a stale pre-migration
snapshot to calico-node for up to 30 minutes, breaking network connectivity for
pods started after the migration.

After a control plane migration, the `Network` resource is annotated with
`networking.calico.extensions.gardener.cloud/typha-migration-restart-at`. This
timestamp is propagated through chart values into the calico-typha pod template,
triggering a rolling restart via the `ManagedResource`. The restart is delayed
until the `WorkloadEndpoint` watch cache returns a non-zero resourceVersion,
ensuring Typha reconnects to a fully warm API server.

Two triggers are handled:
- **CPM (restore):** `Network.Restore` sets the annotation before reconciling.
- **non-HA -> HA:** `Network.Reconcile` detects the transition by comparing
  `cluster.Shoot.Spec.ControlPlane.HighAvailability` against the last-observed
  state stored in `networking.calico.extensions.gardener.cloud/control-plane-ha`.

The annotation value is persisted on the `Network` resource so subsequent
reconciles render the same pod template and do not cause further restarts.

Fixes #

**Release note**:

```other operator
Fixed an issue where calico-typha served stale workload endpoint data for up to
30 minutes after a control plane migration or non-HA to HA transition,
causing pods started after the migration to lose network connectivity.
```


pkg/controller/actuator_restore.go
```diff
@@ -6,13 +6,107 @@ package controller
...
+		// Expected during shoot API server startup after a control plane migration;
+		// the reconciler will requeue until the cache is warm.
+		return fmt.Errorf("shoot API server not yet ready after control plane migration, requeueing: %w", err)
+	}
+
+	patch := client.MergeFrom(network.DeepCopy())
+	if network.Annotations == nil {
+		network.Annotations = map[string]string{}
+	}
+	network.Annotations[restartedAtKey] = time.Now().UTC().Format(time.RFC3339)
```
DockToFuture, line 70: Restore isn't idempotent ... one CPM can trigger multiple typha/calico-node restarts.
Restore ends with return a.Reconcile(...), so a transient error (shoot client not ready, watch cache not warm, chart apply hiccup) requeues and re-enters Restore for the same migration while gardener.cloud/operation: restore is still set. Each pass writes a fresh time.Now() into the ...restarted-at annotation, which changes the pod-template value and rolls the pods again ... so one CPM that retries a few times = several unnecessary restarts in an already-fragile window.

A "only re-annotate if newer" guard doesn't help, since time.Now() is always newer than the value written seconds ago.

I'd track which migration was already handled and make retries no-ops ... the same "record observed state, act only on a change" approach handleHATransition already uses. metadata.generation is a good candidate for that marker: it's stable across the restore retries but changes for a genuinely new resource/migration, so you'd stamp the restart timestamp (and record the handled generation) only when it differs from what's already recorded. WDYT?

pkg/controller/actuator_reconcile.go
```diff
@@ -299,6 +304,54 @@ func (a *actuator) Reconcile(ctx context.Context, log logr.Logger, network *exte
...
+		desiredHAAnnotation = "true"
+	}
+
+	if network.Annotations[calico.AnnotationControlPlaneHA] == desiredHAAnnotation {
+		return nil
+	}
+
+	typhaEnabled, err := isTyphaEnabled(network)
+	if err != nil {
+		log.Error(err, "Failed to decode NetworkConfig for HA transition, skipping Typha restart")
```
DockToFuture: Wouldn't it make sense to return the error here?

pkg/controller/actuator_restore.go
axel7born: Yes, good catch, restore has to be idempotent. 
I think it got lost when I added the restart of the calico-node pods
The generation approach would work, but the additional complexity is actually not needed.
Annotations added by the extension are not carried over. So it's sufficient to check the existence of the restarted-at annotation.

pkg/controller/actuator_reconcile.go
axel7born: Done

DockToFuture (approved):
/lgtm

